PR #5, opened 2022-08-19
Add support for v2 schemas

Bargsteen:
## Purpose

Part of https://github.com/Concordium/concordium-rust-smart-contracts/issues/155

## Changes

- Use v2 schemas when building v1 contracts
- Use schemas on errors when simulating contracts

## Checklist

      hard-to-understand areas.

### DIFF
--- a/cargo-concordium/src/main.rs
+++ b/cargo-concordium/src/main.rs
@@ -477,7 +477,6 @@ fn print_contract_schema_v1(
     }
 }
 
-// TODO: Consider refactoring to reuse v1 version.
 /// Print the contract name and its entrypoints.
 fn print_contract_schema_v2(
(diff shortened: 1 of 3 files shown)


cargo-concordium/src/main.rs
```diff
@@ -471,6 +477,34 @@ fn print_contract_schema_v1(
     }
 }
 
+// TODO: Consider refactoring to reuse v1 version.
```
abizjak, line 480: Let's not have this TODO.

This code is fine and understandable. While there is a bit of duplication, introducing an abstraction to handle the two will also introduce more code. 

Once we have the fourth schema version then perhaps we can refactor :)

abizjak (approved):

PR #7, opened 2022-09-12
Add init templates

### LINKED ISSUE #6: cargo-concordium templates
**Task description**
`cargo-concordium` should have an `init` command to spin up a fresh Concordium smart contract project easily.
`cargo-concordium` should have an `init cis2-nft` command to spin up a fresh Concordium smart contract token project easily.

**Sub-tasks**
- [ ] Add `cargo-concordium init` command
- [ ] Add `cargo-concordium init --template cis2-nft` command

DOBEN:
## Purpose

Add init templates to cargo-concordium

closes #6

## Changes

- [x]  Add `cargo-concordium init` command (the default path is the current folder)
- [x]  Add `cargo-concordium init --path` (creates a new project in the given path; relative or absolute paths are supported)

The templates are pulled from here ( `default` and  `cis2-nft` templates are currently available):
https://github.com/Concordium/concordium-rust-smart-contracts/pull/165

## Checklist

      hard-to-understand areas.


cargo-concordium/src/templates.rs
```diff
@@ -0,0 +1,1179 @@
+#[rustfmt::skip]
```
abizjak, line 1: I think it would be better  to have these as separate files. You can then use the https://doc.rust-lang.org/std/macro.include_bytes.html macro to embed it into the binary.

I think that makes it easier to maintain.

cargo-concordium/src/templates/cis2-nft-cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "cis2_wccd"
+version = "0.1.0"
+authors = ["Concordium <developers@concordium.com>"]
```
abizjak, line 4: Ideally this would not be Concordium, but rather generated from the users' .gitconfig or similar environment variable, or supplied on the command-line.

cargo-concordium/src/templates/cis2-nft-cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "cis2_wccd"
```
abizjak, line 2: Ideally this would be parametrized and the user would name it.

`cargo init` either takes the name of the directory you supply as the crate name, or if you don't supply a directory then it takes the name of the parent directory.

Could we do something similar?

cargo-concordium/src/templates/cis2-nft-lib.rs
```diff
@@ -0,0 +1,1057 @@
+//! A NFT smart contract example using the Concordium Token Standard CIS2.
+//!
+//! # Description
+//! An instance of this smart contract can contain a number of different token
+//! each identified by a token ID. A token is then globally identified by the
+//! contract address together with the token ID.
+//!
```
abizjak, line 7: It seems unsatisfactory to duplicate this code from the examples folder. That's hard to maintain.

I assume this is the same example as the one in rust-smart-contracts?

Could we instead add that repository as a submodule here? Is there some other way we could link to those files instead?

cargo-concordium/src/templates/default-lib.rs
```diff
@@ -0,0 +1,81 @@
+//! # A Concordium V1 smart contract
```
abizjak, line 1: Even if we don't add a submodule, these templates should be moved out of the `src` folder into, e.g., `data` or `resources` folder. `src` should be for sources of the project itself, i.e., the thing that gets compiled.

abizjak (commented):
I think we need to discuss about the approach and maintainability of this.

cargo-concordium/src/build.rs
```diff
@@ -314,6 +314,45 @@ pub fn build_contract_schema<A>(
     Ok(schema)
 }
 
+/// Create a new Concordium smart contract project from a template, or there
+/// are runtime exceptions that are not expected then this function returns
+/// Err(...).
+///
+/// Otherwise a boolean is returned, signifying that the creation was
+/// successful.
```
abizjak, line 322: It looks like this always returns either an error or `Ok(true)`. If that is the case we should remove the boolean and just return `Ok(())`.

cargo-concordium/src/build.rs
```diff
@@ -314,6 +314,45 @@ pub fn build_contract_schema<A>(
     Ok(schema)
 }
 
+/// Create a new Concordium smart contract project from a template, or there
+/// are runtime exceptions that are not expected then this function returns
+/// Err(...).
```
abizjak, line 319:
```suggestion
/// `Err(...)`.
```

cargo-concordium/src/build.rs
```diff
@@ -314,6 +314,45 @@ pub fn build_contract_schema<A>(
...
+        env::current_dir()?.join(path)
+    };
+
+    let result = Command::new("cargo")
+        .arg("generate")
+        .args(&[
+            "--git",
+            "https://github.com/Concordium/concordium-rust-smart-contracts",
+            "--branch",
+            "164-add-default-and-cis2-nft-smart-contract-templates",
```
abizjak, line 338: I suppose this should be changed/removed before we merge this PR.

cargo-concordium/src/build.rs
```diff
@@ -314,6 +314,45 @@ pub fn build_contract_schema<A>(
...
+            "templates",
+        ])
+        .args(&["--destination", absolute_path.to_str().unwrap()])
+        .stdout(Stdio::inherit())
+        .stderr(Stdio::inherit())
+        .stdin(Stdio::inherit())
+        .output()
+        .context("Could not run git clone github_link")?;
```
abizjak, line 346: This error probably won't make too much sense to the user. Maybe something like `Could not obtain the template.`.

cargo-concordium/src/build.rs
```diff
@@ -314,6 +314,45 @@ pub fn build_contract_schema<A>(
...
+
+    if !result.status.success() {
+        anyhow::bail!("Git clone failed.");
+    }
```
abizjak, line 350: Maybe
```suggestion
    ensure!(result.status.success(), "Could not use the template to initialize the project.")
```

cargo-concordium/src/main.rs
```diff
@@ -314,6 +328,14 @@ pub fn main() -> anyhow::Result<()> {
                 build_and_run_wasm_test(&args).context("Could not build and run tests.")?;
             ensure!(success, "Test failed");
         }
+        Command::Init { path } => {
+            let success = init_concordium_project(path)
+                .context("Could not create a new Concordium smart contract project.")?;
+            ensure!(
+                success,
+                "Concordium smart contract project creation failed."
+            );
```
abizjak, line 337: then you probably don't need this

cargo-concordium/Cargo.toml
```diff
@@ -20,6 +20,7 @@ ptree = "0.4"
 strsim = "0.10"
 num-bigint = "0.4.3"
 num-traits = "0.2.15"
+cargo-generate = "0.16.0"
```
abizjak, line 23: I don't see why this is needed. Did you want to use `cargo-generate` as a library?

cargo-concordium/src/build.rs
```diff
@@ -314,6 +314,45 @@ pub fn build_contract_schema<A>(
...
+pub fn init_concordium_project(path: impl AsRef<Path>) -> anyhow::Result<bool> {
+    let path = path.as_ref();
+
+    let absolute_path = if path.is_absolute() {
+        path.to_path_buf()
+    } else {
+        env::current_dir()?.join(path)
+    };
+
+    let result = Command::new("cargo")
```
abizjak, line 332: I expect a common issue will be that users don't have `cargo-generate` installed so we should check if they do. Perhaps using the https://crates.io/crates/which crate.

Something like
```rust
if let Err(which::Error::CannotFindBinaryPath) = which::which("cargo-generate") {
   bail!("`cargo concordium init` requires `cargo-generate` which does not appear to be installed. You can install it by running `cargo install --locked cargo-generate`)
}
```
making sure that the error is displayed in a nice way.
```

File: cargo-concordium/src/main.rs [hunk 14]
```diff
@@ -67,6 +67,20 @@ enum Command {
         )]
         args: Vec<String>,
     },
+    #[structopt(
+        name = "init",
+        about = "Create a new Concordium smart contract project in the existing directory."
```
abizjak, line 72: Can we add to the `help` that this needs `cargo-generate` installed?

Bargsteen:
I had a look, and it looks good in general. It's going to be a great feature! I can take a look again once the remaining comments have been resolved.

Bargsteen (commented):
I had a look, and it looks good in general. It's going to be a great feature! I can take a look again once the remaining comments have been resolved.

cargo-concordium/src/build.rs
DOBEN: yes, the description of this pull request has a note about it:

Once the templates are merged (MR165), we can update the link in cargo-concordium to pull from the main branch rather than MR165.

abizjak: Ah, missed it. Thanks.

abizjak (approved):
This looks good to me. 

Just need to wait to merge the rust-smart-contracts branch and we are golden.

But could you please also add this new command to the README of the project? We have other commands listed.

Bargsteen (approved):
LGTM 👍

PR #15, opened 2022-11-10
Add error messages to output flags when no file name is given

### LINKED ISSUE #13: Cargo-concordium output flags
**Description**

When a user looks up the command for `--out/--schema-out` in `cargo concordium build --help` it says:

>  -o, --out <out>                              Writes the resulting module to file at specified location.
>  -s, --schema-out <schema-out>    Builds the contract schema and writes it to file at specified location.

Users often think that they just have to give a path as follows and the file will be created for them:

> cargo concordium build --out ./
> cargo concordium build --schema-out  ./

but this results in an error as follows
```
Error: Could not build smart contract.
Caused by:
    Is a directory (os error 21)
```

Users have to give the path + file name for this to work e.g.
```
cargo concordium build --out ./my_wasm_file.wasm.v1
cargo concordium build --schema-out  ./schema.bin
```

Suggestion changes: 
- Option 1: Create a default file name in case the user has no file name specified.
- Option 2: Check if the user forgot to specify a file name and give an error message stating how the format to these flags is expected.

DOBEN:
## Purpose

closes #13

## Changes

Add better error messages to provide users more information on how the `--out` and `--schema-out` flags expect a `path + file name` as input.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/cargo-concordium/src/build.rs
+++ b/cargo-concordium/src/build.rs
@@ -175,13 +175,26 @@ pub fn build_contract(
     let data_size = (output_bytes.len() - 8) as u32;
     (&mut output_bytes[4..8]).copy_from_slice(&data_size.to_be_bytes());
 
-    let out_filename = out.unwrap_or_else(|| {
-        let extension = match version {
-            utils::WasmVersion::V0 => "v0",
-            utils::WasmVersion::V1 => "v1",
-        };
-        PathBuf::from(format!("{}.{}", filename, extension))
-    });
+    let out_filename = match out {
+        Some(out) => {
+            // A path name and a filename needs to be provided when using the `--out` flag.
+            if out.file_name().is_none() {
+                anyhow::bail!(
+                    "The `--out` flag requires a path and a filename e.g. \
+                     `./my/path/my_smart_contract.wasm.v1`"
+                );
+            }
+            out
+        }
+        None => {
+            let extension = match version {
+                utils::WasmVersion::V0 => "v0",
+                utils::WasmVersion::V1 => "v1",
+            };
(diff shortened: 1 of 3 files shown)


cargo-concordium/src/build.rs
```diff
@@ -175,13 +175,26 @@ pub fn build_contract(
...
-        let extension = match version {
-            utils::WasmVersion::V0 => "v0",
-            utils::WasmVersion::V1 => "v1",
-        };
-        PathBuf::from(format!("{}.{}", filename, extension))
-    });
+    let out_filename = match out {
+        Some(out) => {
+            // A path name and a filename needs to be provided when using the `--out` flag.
+            if out.file_name().is_none() {
```
abizjak, line 181: Are you sure that this does the right thing? Looking at the documentation https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.file_name

it looks like if the user supplies `/home/user/Desktop` this check will fail, but the tool will still fail to write the schema out properly.

We probably need to additionally test if the user supplied a directory. https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.is_dir

abizjak: Perhaps I misunderstood the intent though.

abizjak (commented):
Please also add the fix to the changelog.

cargo-concordium/src/build.rs
DOBEN: I added the following to the check:
```
   if out.file_name().is_none() || out.is_dir() {
    // throw error that path + file name has to be given
   }
```

abizjak (approved):

PR #26, opened 2023-01-17
Add conversion of schema into its JSON representation

### LINKED ISSUE #25: Get schema for specific entrypoints
**Task description**

Add a new command to get schema types outputted in a JSON file.

E.g.
`cargo concordium schema-expand --module foo.wasm.v1 --out schema-types.json`
`cargo concordium schema-expand --schema schema.bin --out schema-types.json`

The input can be either a smart contract module or a schema. The output should only include the schema types. The purpose is that a dApp should be able to provide the schema types to the browser wallet (e.g. just the schemas to call the `balanceOf` function) without specifying and linking it to a specific smart contract name. E.g. the dApp might want to interact with the `balanceOf` function on some CIS2 token contract (without knowing the exact name of the CIS2 token contract).

DOBEN:
## Purpose

closes #25 

The purpose is that a dApp should be able to provide the schema types to the browser wallet (e.g. just the schemas to call the `balanceOf` function) without specifying and linking it to a specific smart contract name. The dApp might want to interact with the `balanceOf` function on some CIS2 token contract (without knowing the exact name of the CIS2 token contract).

## Changes
Add the commands:
```
cargo concordium schema-json --schema ./schema.bin --out ./my/path/
cargo concordium schema-json --module ./foo.wasm.v1 --out ./my/path/

cargo concordium build --schema-json-out  ./my/path/
````

You need to use exactly one of the two flags(``--schema`` or ``--module``) with this command. The schema for each smart contract will be converted into JSON and then outputted into a separate file e.g. `MyContract1_schema.json` and `MyContract2_schema.json`

## Checklist

      hard-to-understand areas.

### CODE REVIEW

File: cargo-concordium/src/build.rs [hunk 1]
```diff
@@ -375,6 +381,291 @@ pub fn init_concordium_project(path: impl AsRef<Path>) -> anyhow::Result<()> {
...
 
+/// Converts the ContractV0 schema of the given contract_name to JSON and writes
+/// it to a file named after the smart contract name at the specified location.
+pub fn write_json_schema_to_file_v0(
+    mut absolute_path_of_out: PathBuf,
+    contract_name: &str,
+    contract_schema: &ContractV0,
+) -> anyhow::Result<()> {
+    // create empty schema_json
+    let mut schema_json: Value = serde_json::from_str(r#"{}"#)?;
```
abizjak, line 392:
```suggestion
    let mut schema_json: Value = serde_json::Value::Object(Default::default()))
```

cargo-concordium/src/build.rs
```diff
@@ -375,6 +381,291 @@ pub fn init_concordium_project(path: impl AsRef<Path>) -> anyhow::Result<()> {
...
+
+    // add state schema
+    if let Some(state_schema) = &contract_schema.state {
+        schema_json["state"] = json!(general_purpose::STANDARD.encode(to_bytes(state_schema)))
+    }
+
+    // add receive entrypoints
+    if !contract_schema.receive.is_empty() {
+        // create empty entrypoints
+        let mut entrypoints: Value = serde_json::from_str(r#"{}"#)?;
```
abizjak, line 407: Same here.

cargo-concordium/src/build.rs
```diff
@@ -375,6 +381,291 @@ pub fn init_concordium_project(path: impl AsRef<Path>) -> anyhow::Result<()> {
...
+    mut absolute_path_of_out: PathBuf,
+    contract_name: &str,
+    contract_schema: &ContractV0,
+) -> anyhow::Result<()> {
+    // create empty schema_json
+    let mut schema_json: Value = serde_json::from_str(r#"{}"#)?;
+
+    // add init schema
+    if let Some(init_schema) = &contract_schema.init {
+        schema_json["init"] = json!(general_purpose::STANDARD.encode(to_bytes(init_schema)))
```
abizjak, line 396:
```suggestion
        schema_json["init"] = general_purpose::STANDARD.encode(to_bytes(init_schema)).into()
```

cargo-concordium/src/build.rs
```diff
@@ -375,6 +381,291 @@ pub fn init_concordium_project(path: impl AsRef<Path>) -> anyhow::Result<()> {
...
+    }
+
+    // add state schema
+    if let Some(state_schema) = &contract_schema.state {
+        schema_json["state"] = json!(general_purpose::STANDARD.encode(to_bytes(state_schema)))
```
abizjak, line 401:
```suggestion
        schema_json["state"] = general_purpose::STANDARD.encode(to_bytes(state_schema)).into()
```

cargo-concordium/src/build.rs
```diff
@@ -375,6 +381,291 @@ pub fn init_concordium_project(path: impl AsRef<Path>) -> anyhow::Result<()> {
...
+
+            // add parameter schema to receive_function_object
+            if let Some(parameter_schema) = &receive_schema.parameter {
+                receive_function_object["parameter"] =
+                    json!(general_purpose::STANDARD.encode(to_bytes(parameter_schema)))
+            }
+
+            // add return_value schema to receive_function_object
+            if let Some(return_value_schema) = &receive_schema.return_value {
+                receive_function_object["return_value"] =
```
abizjak, line 641: We should use `camelCase` (i.e., `returnValue`) for JSON I think. That's what we do usually.

cargo-concordium/src/main.rs
```diff
@@ -334,6 +375,167 @@ pub fn main() -> anyhow::Result<()> {
...
+                                ))
+                            );
+                        }
+                        schema
+                    }
+                };
+
+                schema
+                    .as_ref()
+                    .map_err(|_| anyhow::anyhow!("Could not deserialize schema file."))?;
```
abizjak, line 442:
```suggestion
                let schema = schema.as_ref().contesxt("Could not deserialize schema file.")?;
```

cargo-concordium/src/main.rs
```diff
@@ -334,6 +375,167 @@ pub fn main() -> anyhow::Result<()> {
...
+
+                match schema.as_ref().unwrap() {
```
abizjak, line 444:
```suggestion
                match schema {
```

cargo-concordium/src/main.rs
```diff
@@ -334,6 +375,167 @@ pub fn main() -> anyhow::Result<()> {
...
+
+            if let Some(schema_path) = schema_path {
+                let bytes = fs::read(schema_path).context("Could not read schema file.")?;
+
+                let schema = if bytes.starts_with(VERSIONED_SCHEMA_MAGIC_HASH) {
+                    from_bytes::<VersionedModuleSchema>(&bytes)
+                } else {
+                    // Question: What is if this is a schema originated from a `WasmVersion::V0`
+                    // module, then this should be 'VersionedModuleSchema::V0', isn't it?
+                    from_bytes(&bytes).map(VersionedModuleSchema::V1)
```
abizjak, line 492: Yes, you have to ask this from the user, as an extra option on the command-line.

cargo-concordium/src/main.rs
```diff
@@ -334,6 +375,167 @@ pub fn main() -> anyhow::Result<()> {
...
+                };
+
+                schema
+                    .as_ref()
+                    .map_err(|_| anyhow::anyhow!("Could not deserialize schema file."))?;
```
abizjak, line 497: Same as above.

cargo-concordium/src/main.rs
```diff
@@ -83,6 +84,54 @@ enum Command {
...
+            required_unless = "module",
+            help = "Path and filename to a file with a schema (expected input: \
+                    `./my/path/schema.bin`)."
+        )]
+        schema_path:  Option<PathBuf>,
+        #[structopt(
+            name = "wasm-version",
+            long = "wasm-version",
+            conflicts_with = "module",
+            help = "If the supplied schema is the legacy unversioned one this should used to \
```
DOBEN, line 119:
```suggestion
            help = "If the supplied schema is the legacy unversioned one this flag should be used to \
```

cargo-concordium/src/main.rs
```diff
@@ -394,10 +512,19 @@ pub fn main() -> anyhow::Result<()> {
...
                     fs::write(schema_out, &module_schema_bytes)
                         .context("Could not write schema file.")?;
                 }
+                if let Some(schema_json_out) = schema_json_out {
+                    ensure!(
+                        schema_json_out.is_dir(),
+                        "The `--schema-expand-out` flag should point to an existing directory \
+                         (expected input `./my/path`)."
+                    );
+                    eprintln!("   Writing schema to {}.", schema_json_out.display());
```
DOBEN, line 524: Not sure if we need it since we also print the below message in the `build.rs` file, but also don't mind if we print out too much here :)

    println!(
        "   Writing schema for {} to {}.",
        contract_name,
        out_path.display()
    );

DOBEN (commented):
Well done.

cargo-concordium/src/main.rs
```diff
@@ -394,10 +512,19 @@ pub fn main() -> anyhow::Result<()> {
                         );
                     }
 
-                    eprintln!("   Writing schema to {}.", schema_out.display());
                     fs::write(schema_out, &module_schema_bytes)
                         .context("Could not write schema file.")?;
                 }
+                if let Some(schema_json_out) = schema_json_out {
+                    ensure!(
+                        schema_json_out.is_dir(),
```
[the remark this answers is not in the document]
abizjak: I expected that `is_dir` implies `!is_file`. But I have not tested this. If it does not then I agree.

cargo-concordium/src/main.rs
```diff
@@ -394,10 +512,19 @@ pub fn main() -> anyhow::Result<()> {
...
+                        "The `--schema-expand-out` flag should point to an existing directory \
+                         (expected input `./my/path`)."
```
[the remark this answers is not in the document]
abizjak: That's a good point.

cargo-concordium/src/main.rs
abizjak: I agree, this should be removed. I removed it in the other place already, but forgot this.

cargo-concordium/src/build.rs
```diff
@@ -375,6 +384,237 @@ pub fn init_concordium_project(path: impl AsRef<Path>) -> anyhow::Result<()> {
...
+    let mut out_path = absolute_path_of_out.to_owned();
+    out_path.push(contract_name.to_owned() + "_schema.json");
+
+    println!(
+        "   Writing schema for {} to {}.",
+        contract_name,
+        out_path.display()
+    );
+    std::fs::write(out_path, serde_json::to_string_pretty(&schema_json)?)
+        .context("Unable to write schema output.")?;
```
limemloh, line 432: You should add context to `to_string_pretty` here

Also, instead of allocating a string and then writing that string to file, you can write directly to the file using something like this
```suggestion
    let out_file = std::fs::File::create(out_path).context(..)?;
    serde_json::to_writer_pretty(out_file, &schema_json)
        .context("Unable to write schema output.")?;
```

cargo-concordium/src/main.rs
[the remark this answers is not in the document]
abizjak: At least on linux `is_dir` seems to imply `!is_file`. So while the test you suggest is also correct, it does not seem necessary.

cargo-concordium/src/main.rs
```diff
@@ -83,6 +84,54 @@ enum Command {
...
+            conflicts_with = "module",
+            required_unless = "module",
+            help = "Path and filename to a file with a schema (expected input: \
+                    `./my/path/schema.bin`)."
+        )]
+        schema_path:  Option<PathBuf>,
+        #[structopt(
+            name = "wasm-version",
+            long = "wasm-version",
+            conflicts_with = "module",
```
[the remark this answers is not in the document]
abizjak: I removed this alltogether and now use it both when supplying the schema, and the module. In case somebody supplies the old module without an explicit version.

cargo-concordium/src/build.rs
```diff
@@ -375,6 +384,237 @@ pub fn init_concordium_project(path: impl AsRef<Path>) -> anyhow::Result<()> {
...
+            entrypoints[method_name] = type_to_json(receive_schema);
+        }
+
+        // add all receive entrypoints
+        schema_json["entrypoints"] = entrypoints;
+    }
+
+    // save the schema JSON representation into the file.
+    let mut out_path = absolute_path_of_out.to_owned();
+    out_path.push(contract_name.to_owned() + "_schema.json");
```
limemloh, line 424: Consider sanitizing the contract name, since a contract name might contain `/` or characters which are not allowed in file names.

cargo-concordium/src/build.rs
```diff
@@ -375,6 +384,237 @@ pub fn init_concordium_project(path: impl AsRef<Path>) -> anyhow::Result<()> {
...
+
+    println!(
+        "   Writing schema for {} to {}.",
+        contract_name,
+        out_path.display()
+    );
+    std::fs::write(out_path, serde_json::to_string_pretty(&schema_json)?)
+        .context("Unable to write schema output.")?;
+
+    Ok(())
```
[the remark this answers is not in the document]
abizjak: I don't think that adds much value.

cargo-concordium/src/build.rs
abizjak: It is not clear that writing to the writer is better. Usually you then have to make a buffered writer first from the file. Otherwise it is worse.

DOBEN:
I get this error when running cargo concordium (we are using twice the `short = "s"` flag):

thread 'main' panicked at 'Argument short must be unique

	-s is already in use', ...../src/app/parser.rs:190:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

cargo-concordium/src/build.rs
abizjak: only ascii alphanumeric and `"-_[]{}"` are now allowed.

cargo-concordium/src/build.rs
abizjak: to_string_pretty does not really fail so there is no value adding context to it.

cargo-concordium/src/main.rs
```diff
@@ -394,10 +514,18 @@ pub fn main() -> anyhow::Result<()> {
...
                     }
 
-                    eprintln!("   Writing schema to {}.", schema_out.display());
                     fs::write(schema_out, &module_schema_bytes)
                         .context("Could not write schema file.")?;
                 }
+                if let Some(schema_json_out) = schema_json_out {
+                    ensure!(
+                        schema_json_out.is_dir(),
+                        "The `--schema-expand-out` flag should point to an existing directory \
```
DOBEN, line 523:
```suggestion
                        "The `--schema-json-out` flag should point to an existing directory \
```

cargo-concordium/src/main.rs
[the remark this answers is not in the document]
DOBEN: yes, I think you are right. I was not able to create an edge example that the simple check `schema_json_out.is_dir()` was not sufficient.

cargo-concordium/src/main.rs
```diff
@@ -334,16 +391,77 @@ pub fn main() -> anyhow::Result<()> {
...
         }
+        Command::SchemaJSON {
+            out,
+            module_path,
+            schema_path,
+            wasm_version,
+        } => {
+            // A valid path needs to be provided when using the `--out` flag.
+            ensure!(
+                out.is_dir() && !out.is_file(),
```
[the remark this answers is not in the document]
DOBEN: yes, I think you are right. I was not able to create an edge example that the simple check out.is_dir() was not sufficient.

cargo-concordium/src/main.rs
```diff
@@ -334,16 +391,77 @@ pub fn main() -> anyhow::Result<()> {
...
+                    match wv {
+                        WasmVersion::V0 => from_bytes(&bytes).map(VersionedModuleSchema::V0)?,
+                        WasmVersion::V1 => from_bytes(&bytes).map(VersionedModuleSchema::V1)?,
+                    }
+                } else {
+                    bail!(
+                        "Legacy unversioned schema was supplied, but no version was provided. Use \
+                         `--wasm-version` to specify the version."
+                    );
+                }
```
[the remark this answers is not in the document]
abizjak: I will not do this refactoring now

cargo-concordium/src/main.rs
```diff
@@ -334,16 +391,77 @@ pub fn main() -> anyhow::Result<()> {
...
+            let schema = if let Some(module_path) = module_path {
+                let bytes = fs::read(module_path).context("Could not read module file.")?;
+
+                let mut cursor = std::io::Cursor::new(&bytes[..]);
+                let wasm_version = utils::WasmVersion::read(&mut cursor)
+                    .context("Could not read module version from the supplied module file.")?;
+
+                match wasm_version {
+                    utils::WasmVersion::V0 => {
+                        let module = &cursor.into_inner()[8..];
```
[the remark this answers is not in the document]
abizjak: Moving it out changes the logic. Before the match we don't know if bytes has length at least 8 or not.

abizjak:
> I get this error when running cargo concordium (we are using twice the `short = "s"` flag):
> 
> thread 'main' panicked at 'Argument short must be unique
> 
> ```
> -s is already in use', ...../src/app/parser.rs:190:13
> ```
> 
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

Good catch. This works now

```
cargo concordium build -e -s schema/schema.bin -j schema -o memo.wasm.v1
   Compiling crunchy v0.2.2
   Compiling static_assertions v1.1.0
   Compiling hex v0.4.3
   Compiling byteorder v1.4.3
   Compiling concordium-contracts-common v5.1.1 (/home/abizjak/Private/Documents/Concordium/concordium-rust-smart-contracts/concordium-contracts-common/concordium-contracts-common)
   Compiling fixed-hash v0.7.0
   Compiling uint v0.9.3
   Compiling primitive-types v0.11.1
   Compiling concordium-std-derive v5.1.0 (/home/abizjak/Private/Documents/Concordium/concordium-rust-smart-contracts/concordium-std-derive)
   Compiling concordium-std v5.1.0 (/home/abizjak/Private/Documents/Concordium/concordium-rust-smart-contracts/concordium-std)
   Compiling concordium-cis2 v2.0.0 (/home/abizjak/Private/Documents/Concordium/concordium-rust-smart-contracts/concordium-cis2)
   Compiling cis2_nft v0.1.0 (/home/abizjak/Private/Documents/Concordium/concordium-rust-smart-contracts/examples/cis2-nft)
    Finished release [optimized] target(s) in 3.26s
   Compiling concordium-std-derive v5.1.0 (/home/abizjak/Private/Documents/Concordium/concordium-rust-smart-contracts/concordium-std-derive)
   Compiling concordium-std v5.1.0 (/home/abizjak/Private/Documents/Concordium/concordium-rust-smart-contracts/concordium-std)
   Compiling concordium-cis2 v2.0.0 (/home/abizjak/Private/Documents/Concordium/concordium-rust-smart-contracts/concordium-cis2)
   Compiling cis2_nft v0.1.0 (/home/abizjak/Private/Documents/Concordium/concordium-rust-smart-contracts/examples/cis2-nft)
    Finished release [optimized] target(s) in 2.35s

   Module schema includes:

     Contract schema: 'cis2_nft' in total 2957 B.
       event   : 598 B
       receive
        - 'balanceOf'       : 258 B
        - 'mint'            : 247 B
        - 'operatorOf'      : 289 B
        - 'setImplementors' : 211 B
        - 'supports'        : 236 B
        - 'tokenMetadata'   : 240 B
        - 'transfer'        : 321 B
        - 'updateOperator'  : 272 B
        - 'view'            : 158 B

     Contract schema: 'cis2_nft/' in total 604 B.
       event   : 598 B

   Total size of the module schema is 3593 B
   Writing schema for cis2_nft to schema/cis2_nft_schema.json.
   Writing schema for cis2_nft/ to schema/contract-schema_1.json.
   Embedding schema into module.

    Finished smart contract module 82.895 kB
```

abizjak (approved):

cargo-concordium/src/build.rs
```diff
@@ -375,6 +384,228 @@ pub fn init_concordium_project(path: impl AsRef<Path>) -> anyhow::Result<()> {
...
+    let mut out_path = root.to_path_buf();
+
+    // Make sure the path is valid on all platforms.
+    let file_name = if contract_name
+        .chars()
+        .all(|c| c.is_ascii_alphanumeric() || "-_[]{}".contains(c))
+    {
+        contract_name.to_owned() + "_schema.json"
+    } else {
+        format!("contract-schema_{}.json", counter)
```
Bargsteen, line 407: Perhaps we should print a warning when the provided name isn't used.

Bargsteen (commented):
Nice work 😊

PR #39, opened 2023-03-29
Misc changes

abizjak:
## Purpose

In lieu of review of #23 

## Changes 

- Remove the need to clone all accounts, contracts, modules when executing contracts. Only references are passed in.
- Make contract_invoke require an immutable reference only.
- Avoid mutating the state of an account before execution. Instead remember which account is being updated for correct balance queries. This is cleaner and makes it simpler to enforce transactionality since any errors in execution will automatically not update the Chain structure.
- Hide internals of the Chain struct, instead expose the relevant accessors.
- Fixes in a number of documentation links.
- Misc improvements.

TODO
- [ ] The next_contract_index as it is now is not ideal. With hidden fields of the Chain it is OK, since initializing is the only way to add new contracts. But in this case we don't need it either.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-smart-contract-testing/src/impls.rs
+++ b/concordium-smart-contract-testing/src/impls.rs
@@ -853,8 +853,8 @@ impl Chain {
     ///     Some(Amount::from_ccd(123))
     /// );
     /// ```
-    pub fn create_account(&mut self, address: AccountAddress, account: Account) -> Option<Account> {
-        self.accounts.insert(address.into(), account)
+    pub fn create_account(&mut self, account: Account) -> Option<Account> {
+        self.accounts.insert(account.address.into(), account)
     }
 
     /// Create a contract address by giving it the next available index.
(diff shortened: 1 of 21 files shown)


concordium-smart-contract-testing/src/impls.rs
```diff
@@ -612,20 +680,11 @@ impl Chain {
...
-                ),
-            });
-        }
         // Ensure the sender exists.
         if !self.address_exists(sender) {
-            // TODO: Should we charge the header cost if the invoker exists but the sender
-            // doesn't?
+            // This situation never happens on the chain since to send a message the sender
+            // is verifier upfront. So what we do here is custom behaviour, and we reject
+            // without consuming any eneryg.
```
Bargsteen, line 687:
```suggestion
            // is verified upfront. So what we do here is custom behavior, and we reject
            // without consuming any energy.
```

concordium-smart-contract-testing/src/impls.rs
```diff
@@ -651,57 +710,86 @@ impl Chain {
...
                 energy_used,
-                transaction_fee: self.calculate_energy_cost(energy_used),
+                transaction_fee: self.parameters.calculate_energy_cost(energy_used),
                 kind: ContractInvokeErrorKind::InsufficientFunds,
             });
         }
 
         // Charge account for the reserved energy up front. This is to ensure that
         // contract queries for the invoker balance are correct.
         // The `amount` is handled in contract_invocation_worker.
```
Bargsteen, line 739:
```suggestion
```

Bargsteen (approved):

PR #40, opened 2023-03-31
Use cargo metadata in cargo-concordium instead of parsing manifest

limemloh:
## Purpose

Use `cargo_metadata` in `cargo-concordium` instead of parsing package manifest.

Commands such as `cargo concordium build` can now be called from a project subdirectory, where before it was required to be called in the same directory as the `Cargo.toml`.

Additionally, `cargo-concordium` uses the target directory from cargo metadata, instead of the hard-coded `target`.

## Changes

**cargo-concordium**

- Replace `cargo_toml` dependency with `cargo_metadata`.
- Remove patch versions of dependencies in Cargo.toml.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/cargo-concordium/CHANGELOG.md
+++ b/cargo-concordium/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Changelog
 
+- Support calling `cargo concordium build` and `cargo concordium test` from a project subdirectory.
+
 ## 2.7.0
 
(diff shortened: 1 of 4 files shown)


abizjak (approved):

DOBEN (approved):
Works well.

PR #42, opened 2023-04-09
Simplifications and helpers

abizjak:
## Purpose

Remove the "logs/events" since last interrupt. That semantics is awkward and I don't think it brings much value.

The entire log can be recovered from the trace_elements.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-smart-contract-testing/src/invocation/types.rs
+++ b/concordium-smart-contract-testing/src/invocation/types.rs
@@ -7,15 +7,8 @@ use concordium_base::{
     },
     smart_contracts::OwnedParameter,
 };
-use concordium_smart_contract_engine::v1::{trie::MutableState, InvokeResponse};
+use concordium_smart_contract_engine::v1::trie::MutableState;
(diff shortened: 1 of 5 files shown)


concordium-smart-contract-testing/src/invocation/types.rs
```diff
@@ -7,20 +7,14 @@ use concordium_base::{
...
-use concordium_smart_contract_engine::{
-    v0,
-    v1::{trie::MutableState, InvokeResponse},
-};
+use concordium_smart_contract_engine::v1::{trie::MutableState, InvokeResponse};
 use std::collections::BTreeMap;
 
 /// The response from invoking an entrypoint.
 #[derive(Debug)]
 pub(crate) struct InvokeEntrypointResponse {
```
Bargsteen, line 15: We should remove this wrapper type, now that there is only one field.

Bargsteen (changes requested):
Looks good, but we should remove the wrapper.

PR #43, opened 2023-04-11
Update base and make corresponding changes in cargo-concordium

Bargsteen:
## Purpose

Update base and make the necessary changes to cargo-concordium for it to build.

## Changes

- Update base
- Fix trivial issues
- Change `get_parameter` to return an `OwnedParameter` instead as it seemed more appropriate.
   - We might consider checking the size of the parameter as well, but that would be a breaking change.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/cargo-concordium/Cargo.toml
+++ b/cargo-concordium/Cargo.toml
@@ -22,8 +22,6 @@ which = "4.3"
 rand = { version = "=0.7", features = ["small_rng"] }
 cargo_metadata = "0.15"
 
-[features]
-
 [dependencies.concordium-wasm]
 path = "../concordium-base/smart-contracts/wasm-transform"
 version = "1.0"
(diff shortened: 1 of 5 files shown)


cargo-concordium/Cargo.toml
```diff
@@ -22,19 +22,21 @@ strsim = "0.10"
 which = "4.3.0"
 rand = { version = "=0.7", features = ["small_rng"] }
 
-[dependencies.wasm-transform]
+[features]
```
abizjak, line 25: Why do we need `features`?

abizjak (approved):

cargo-concordium/Cargo.toml
Bargsteen: Not sure how that got in there 🤷

PR #46, opened 2023-04-18
Vscode extension: build smart contract

### LINKED ISSUE #34: Provide command for building a smart contract in VS Code extension
The extension should provide commands for running `cargo concordium build`

- [x] Variant, which builds with: No schema.
- [x] Variant, which builds with: With Schema (Base64 version of schema should be printed).
- [x] Variant, which builds with: With embedded schema (Base64 version of schema should be printed).
- [x] We should also provide a setting for additional arguments.
- [x] When running this command, it should ensure the `wasm32-unknown-unknown` target is installed, otherwise prompt the user whether to install it.

### LINKED ISSUE #35: Provide command for testing a smart contract in VS Code extension
The extension should provide commands for triggering `cargo concordium test`.

We should also provide a setting for additional arguments.

When running this command, it should ensure the `wasm32-unknown-unknown` target is installed, otherwise prompt the user whether to install it.

limemloh:
## Purpose

Allow for building and testing smart contract projects from within VS Code.
Closes #34 and closes #35.

## Changes

- Add commands for building and testing the smart contract project currently being focused on an editor. Each with a setting for adding additional arguments.
- [Add task provider](https://code.visualstudio.com/api/extension-guides/task-provider) which searches the current workspace for suggestions to build tasks and test tasks.
- Improve error handling for the version command.

## Checklist

      hard-to-understand areas.


vscode-smart-contracts/src/commands.ts
```diff
@@ -15,10 +19,155 @@ export async function version() {
...
+  const schemaArgs =
+    schemaSettings === "skip"
+      ? []
+      : schemaSettings === "embed"
+      ? ["--schema-embed"]
+      : ["--schema-base64-out", "-"];
+  const additionalArgs = config.getAdditionalBuildArgs();
+  const args = schemaArgs.concat(additionalArgs);
+  return vscode.tasks.executeTask(
+    await cargoConcordium.build(cwd, workspaceFolder, args)
```
abizjak, line 94: I feel like we should provide some sort of feedback on success.

Like tell them "your module is built in ...". If I see correctly we don't set any output path, so their module is available in an odd location. It would be better to allow them to select this.

vscode-smart-contracts/src/commands.ts
```diff
@@ -15,10 +19,155 @@ export async function version() {
...
+    editor.document.uri
+  );
+  const schemaArgs =
+    schemaSettings === "skip"
+      ? []
+      : schemaSettings === "embed"
+      ? ["--schema-embed"]
+      : ["--schema-base64-out", "-"];
+  const additionalArgs = config.getAdditionalBuildArgs();
+  const args = schemaArgs.concat(additionalArgs);
```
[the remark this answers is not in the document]
limemloh: So whenever the command is executed, it will launch a VS Code task, which pops up a panel with the output.
Meaning, if they provide invalid arguments, they will see whatever error `cargo-concordium` outputs.

vscode-smart-contracts/src/commands.ts
limemloh: They will see a panel with the output from cargo-concordium, so they are provided feedback.

Regarding the odd location of the output. I have considered this issue as well, but not sure that I would like to prompt the user for the location every time they run the command.
The solution I went with was having them put it as an additional argument, since they could define a workspace config with this, this is not ideal either.

vscode-smart-contracts/src/commands.ts
```diff
@@ -15,10 +19,155 @@ export async function version() {
...
+  const cwd = path.dirname(editor.document.uri.fsPath);
+  const workspaceFolder = vscode.workspace.getWorkspaceFolder(
+    editor.document.uri
+  );
+  const schemaArgs =
+    schemaSettings === "skip"
+      ? []
+      : schemaSettings === "embed"
+      ? ["--schema-embed"]
+      : ["--schema-base64-out", "-"];
```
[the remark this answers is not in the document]
limemloh: Right. I tried this at first, but this schema does not support outputting to stdout.
So I will fix that first.

vscode-smart-contracts/src/commands.ts
```diff
@@ -15,10 +19,155 @@ export async function version() {
...
+ * Embedding the schema into the resulting smart contract.
+ */
+export function buildEmbedSchema(editor: vscode.TextEditor) {
+  return buildWorker(editor, "embed");
+}
+
+/**
+ * Type representing the different settings for schema generation during cargo-concordium build.
+ */
+type SchemaSettings = "skip" | "embed" | "stdout";
```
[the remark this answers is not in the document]
limemloh: Because that is the case for the commands that I chose to define.

Do you want another combination of schema options?

vscode-smart-contracts/src/extension.ts
```diff
@@ -1,21 +1,132 @@
...
+        vscode.Uri.file(cwd)
+      );
+      const additionalArgs = definition.args ?? [];
+      const args = ["--schema-base64-out", "-"].concat(additionalArgs);
+      const resolvedTask = await cargoConcordium.build(
+        cwd,
+        workspaceFolder,
+        args
+      );
+      // resolveTask requires that the same definition object to be used.
```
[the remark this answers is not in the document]
limemloh: I have added a bit to this. Let me know it this is still unclear

vscode-smart-contracts/src/commands.ts
abizjak: I think it would be better to have some default in the root of the package. Named after the package.

That's at least what I usually do when building contracts and it is very easily discoverable.

vscode-smart-contracts/src/commands.ts
```diff
@@ -15,10 +19,155 @@ export async function version() {
...
+    "Build smart contract",
+    execution
+  );
+  return executeAndAwaitTask(task);
+}
+
+/**
+ * Execute and await a task to end.
+ * @param task The task to execute.
+ * @returns Promise which resolves when the task have ended.
```
DOBEN, line 129:
```suggestion
 * @returns Promise which resolves when the task has ended.
```

vscode-smart-contracts/src/commands.ts
```diff
@@ -15,10 +19,155 @@ export async function version() {
...
+ * Check the rustup list of installed targets for wasm32-unknown-unknown.
+ */
+async function haveWasmTargetInstalled() {
+  const out = await exec("rustup target list --installed");
+  return out.stdout.includes("wasm32-unknown-unknown");
+}
+
+/**
+ * Install the wasm32-unknown-unknown target using rustup.
+ * The returned promise resolves when the install task have ended.
```
DOBEN, line 108:
```suggestion
 * The returned promise resolves when the install task has ended.
```

vscode-smart-contracts/src/commands.ts
```diff
@@ -15,10 +19,155 @@ export async function version() {
...
+      "Abort"
+    );
+    if (response === "Abort") {
+      vscode.window.showErrorMessage(
+        "Unable to run tests because of missing wasm32-unknown-unknown target."
+      );
+      return;
+    }
+    await installWasmTarget();
+  }
```
DOBEN, line 163: Nitpicking: Since you check (and prompt the user to install) the correct wasm target in the `test` and `build` functions, this part could be in a separate function so we don't duplicate code.

Line 150-163

vscode-smart-contracts/src/commands.ts
```diff
@@ -15,10 +19,155 @@ export async function version() {
...
+    } catch (e) {
+      reject(e);
+    }
+  });
+}
+
+/**
+ * Run 'cargo-concordium test' using the directory of the currently focused editor.
+ */
+export async function test(editor: vscode.TextEditor) {
```
DOBEN, line 149: Nitpicking: You call the corresponding function `buildWorker` and this function `test`. Is there a difference or we can name it consistently?

vscode-smart-contracts/src/commands.ts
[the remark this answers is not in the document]
limemloh: This was discussed offline, the conclusion is to output the schema to some default output directory instead of promoting a workflow where the user have to copy paste from the stdout

vscode-smart-contracts/src/commands.ts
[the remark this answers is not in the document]
limemloh: done

vscode-smart-contracts/src/commands.ts
limemloh: So now it will output schema and module to an `out` directory. Currently, the resulting module is just named `module.wasm.v1`.

vscode-smart-contracts/src/commands.ts
limemloh: The point of the `Worker` suffix is to indicate this is an internal function which does the main work, since the different build variants are similar. Test on the other hand is exported.

vscode-smart-contracts/src/commands.ts
limemloh: Right, so I did consider this, but I find the resulting code less readable. Mainly because of having to propagate the return early from an inner function.

vscode-smart-contracts/README.md
```diff
@@ -24,12 +36,11 @@ This extension contributes the following settings:
...
+* `concordium-smart-contracts.additional-test-args`: `string[]` (default `[]`) <br>
+  Provide additional arguments for `cargo-concordium` when running the test smart contract command.
 
-<!--
-## Known Issues
-
-Calling out known issues can help limit users opening duplicate issues against your extension.
--->
 
 ## Release Notes
```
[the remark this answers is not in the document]
limemloh: Adding empty directories to git is a bit hacky, I feel this is not too much to ask of a developer.

vscode-smart-contracts/README.md
```diff
@@ -24,12 +36,11 @@ This extension contributes the following settings:
...
+* `concordium-smart-contracts.additional-build-args`: `string[]` (default `[]`) <br>
+  Provide additional arguments for `cargo-concordium` when running the build smart contract command.
+* `concordium-smart-contracts.additional-test-args`: `string[]` (default `[]`) <br>
+  Provide additional arguments for `cargo-concordium` when running the test smart contract command.
 
-<!--
-## Known Issues
-
-Calling out known issues can help limit users opening duplicate issues against your extension.
--->
```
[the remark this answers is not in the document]
limemloh: Alternatively we could add some npm script for this step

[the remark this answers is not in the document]
abizjak: Why does it have to be a npm script? 

Also we should not be distributing the debug version.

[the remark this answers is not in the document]
limemloh: I have added some instructions on this.

vscode-smart-contracts/src/extension.ts
```diff
@@ -1,21 +1,141 @@
...
+          })
+        );
+      })
+    );
+    return tasks.flat();
+  },
+  /**
+   * Resolve a user provided task definition.
+   * Returning undefined fallback to search the output of `provideTasks`
+   * for a matching task, which is slow.
```
abizjak, line 83: I'm confused by this phrasing. Is this function returning an "undefined"? 

What exactly is slow, and how does returning undefined help?

vscode-smart-contracts/src/extension.ts
```diff
@@ -1,21 +1,141 @@
...
+  /** Search the current workspace for possible tasks */
+  async provideTasks() {
+    const workspaceFolders = vscode.workspace.workspaceFolders ?? [];
+    const tasks = await Promise.all(
+      workspaceFolders.map(async (workspaceFolder) => {
+        const workspaceRoot = workspaceFolder.uri.fsPath;
+        const cargoProjectDirs = await getCargoProjectDirs(workspaceRoot);
+        return Promise.all(
+          cargoProjectDirs.flatMap((cwd) => {
+            const defaultOutDir = path.join(cwd, "out");
```
abizjak, line 63: Have you considered using `concordium-out` as the name instead? To make it less likely to clash?

abizjak (approved):

vscode-smart-contracts/src/extension.ts
limemloh: clash with what? this will be in the contract directory, so it seems fine to assume mainly cargo stuff is here

abizjak: I often create "out" directories for stuff. It's such a generic name.

vscode-smart-contracts/package.json
```diff
@@ -23,6 +23,50 @@
...
+        "command": "concordium-smart-contracts.build",
+        "title": "Concordium Smart Contracts: Build contract"
+      },
+      {
+        "command": "concordium-smart-contracts.build-skip-schema",
+        "title": "Concordium Smart Contracts: Build contract (no schema)"
+      },
+      {
+        "command": "concordium-smart-contracts.build-embed-schema",
+        "title": "Concordium Smart Contracts: Build contract (embed schema)"
```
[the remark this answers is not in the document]
DOBEN: I would say that at least from a front-end development perspective, it is still the `base64` schema that is easiest (most convenient) to use currently with our tooling. I have just one schema variable (with the `base64` schema) at the front end that can be fed into every `serialization/invoke/sendTransaction` helper function or the browser wallet and it sorts out for me which schema I need from the blob. 

For example, while still in the active smart contract development phase where the schema changes often, it is much easier to just update one `base64` variable at the front-end, while `json` schema might need more adjustments.

While `json` schema, I think is the better way to represent the schema because it is human-readable (and I would prefer to have the `json` schema completely decoded in a human-readable manner). Currently, it is a bit annoying that our tooling needs to get exactly fed in which part of the `json` schema blob it should use. So my suggestion is actually more unrelated to this pull request but update our front-end tooling that it get the whole `json` schema blob fed into the `serialization/invoke/sendTransaction/browserWallet` and then our tooling sort out based on the entrypoint name which of the schema in the blob is appropriate to use.

vscode-smart-contracts/src/extension.ts
limemloh: I have tried to rephrase this, let me know it is still not clear what returning undefined means

DOBEN (approved):
Looks good to me.

vscode-smart-contracts/src/extension.ts
limemloh: Okay, I'll use `concordium-out` then

PR #48, opened 2023-04-19
cargo-concordium: Create directory path for produced artifacts

limemloh:
## Purpose/Changes

`cargo-concordium` now creates non-existing directories for out paths instead of producing an error.
This is relevant for:

**`cargo concordium build`, when providing:**

- `--out path`
- `--schema-out path`
- `--schema-json-out path`
- `--schema-base64-out path`

**`cargo concordium run update` and `cargo concordium run init` when providing:**
- `--out-bin path`
- `--out-json path`

## Checklist

      hard-to-understand areas.

### DIFF
--- a/cargo-concordium/src/build.rs
+++ b/cargo-concordium/src/build.rs
@@ -216,6 +216,10 @@ pub fn build_contract(
     };
 
     let total_module_len = output_bytes.len();
+    if let Some(out_dir) = out_filename.parent() {
+        fs::create_dir_all(out_dir)
+            .context("Unable to create directory for the resulting smart contract module")?;
+    }
     fs::write(out_filename, output_bytes)?;
     Ok((total_module_len, return_schema))
 }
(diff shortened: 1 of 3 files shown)


cargo-concordium/src/build.rs
```diff
@@ -216,6 +216,10 @@ pub fn build_contract(
     };
 
     let total_module_len = output_bytes.len();
+    if let Some(out_dir) = out_filename.parent() {
+        fs::create_dir_all(out_dir)
+            .context("Unable to create directory for the resulting smart contract module")?;
```
abizjak, line 221: Is there a reason why these messages don't end with a `.`?

abizjak: The other messages we have do.

abizjak (approved):

cargo-concordium/src/build.rs
limemloh: Will fix

PR #51, opened 2023-04-21
Add vscode command for cargo concordium init

### LINKED ISSUE #36: Provide command for smart contract templates in VS Code extension
The extension should provide a command for using `cargo concordium init`, which sets up a template project from a list of options.

Ideally, it would prompt the user for which template to use, but a simpler approach might be to just provide the default.

When running this command, it should ensure `cargo-generate` is installed, otherwise prompt the user whether to install it.

limemloh:
## Purpose

Add command which prompts the user for a directory to launches a terminal with `cargo concordium init`.
Closes #36 

## Checklist

      hard-to-understand areas.


vscode-smart-contracts/package.json
```diff
@@ -42,6 +42,10 @@
       {
         "command": "concordium-smart-contracts.test",
         "title": "Concordium Smart Contracts: Test contract"
+      },
+      {
+        "command": "concordium-smart-contracts.setup-project",
+        "title": "Concordium Smart Contracts: Setup new smart contract project"
```
abizjak, line 48: Is using "Setup" as the wording the common thing to do for this sort of command?

Why not use "initialize" like the cargo command?

abizjak (approved):

vscode-smart-contracts/package.json
limemloh: No particular reason. I'll change this, since it makes sense to follow the same naming

DOBEN (approved):
LGTM

PR #53, opened 2023-04-24
Misc bugfixes

abizjak:
## Purpose

Some bugfixes.

- Policies have a specific serialization so that they can be handled easily inside contracts.
- We need to start a new generation always on initial invocation of the receive method. This ensures that no stale data can be accessed from a previous generation inside the same transaction (stale data would be an entry that was looked up, or an iterator that was alive).
- Use a global (per transaction) modification index. This is how it is in the node as well.

## Checklist

      hard-to-understand areas.


concordium-smart-contract-testing/src/invocation/impls.rs
```diff
@@ -292,6 +313,9 @@ impl<'a, 'b> EntrypointInvocationHandler<'a, 'b> {
                                     // Update the state field with the newest value from the
                                     // changeset.
                                     data.state = self.contract_state(data.address);
+                                    // TODO: This change fixes some costs, but maybe should not be
+                                    // necessary.
+                                    data.state = self.contract_state(data.address);
```
abizjak, line 318: This is redundant.

Bargsteen:
Could you take a look at the latest changes I've made @abizjak? Then we can merge this in.

abizjak:
> Could you take a look at the latest changes I've made @abizjak? Then we can merge this in.

Apologies for delay. I think it looks good, so merging it in.

PR #54, opened 2023-04-24
Vscode prepare publishing

### LINKED ISSUE #52: Automate the build for vscode extension

limemloh:
## Purpose

Setup build job for packaging the VS Code extension.
Closes #52.

## Changes

- Fix broken link in root readme.
- Add extension icon.
- Remove unused dependency `request-light`.
- Added Jenkins files for each platform.
- Ensure extension get activated when concordium task type is needed.

## Checklist

      hard-to-understand areas.


vscode-smart-contracts/scripts/linux-vscode-smart-contracts.Jenkinsfile
```diff
@@ -0,0 +1,72 @@
...
+    environment {
+        VERSION = sh(
+            returnStdout: true,
+            script: '''\
+                # Extract version number if not set as parameter
+                [ -z "$VERSION" ] && VERSION=$(node -p "require('vscode-smart-contracts/package.json').version")
+                echo -n "$VERSION"
+            '''.stripIndent()
+        )
+        CARGO_CONCORDIUM_VERSION = "${VERSION}"
```
abizjak, line 12: I am not sure I understand the purpose of this. The VERSION is extracted from package.json, but now set as the CARGO_CONCORDIUM version.

vscode-smart-contracts/scripts/linux-vscode-smart-contracts.Jenkinsfile
```diff
@@ -0,0 +1,72 @@
...
+                '''.stripIndent()
+            }
+        }
+        stage('build') {
+            agent {
+                docker {
+                    reuseNode true
+                    image 'concordium/base:latest'
+                    args '-u root'
+                }
```
abizjak, line 34: Why are we doing this? I don't think we should use the concordium/base image. That is an odd image. We are building a pure node thing, are we not? Isn't there an official image for that?

abizjak (commented):
Looks good overall.

It looks like you could just parametrize this by the location of the binary for the relevant platform, and just have the one job build everything.

Or does the extension build step depend on the platform as well?

abizjak:
@Thahara I asked for your review as well. If you can do a pass of the Jenkinsfile to make sure it looks reasonable.

vscode-smart-contracts/scripts/linux-vscode-smart-contracts.Jenkinsfile
limemloh: Right, this is a bit weird and the only point here was to provide some sort of default for the variable.

Thahara: I'm guessing `CARGO_CONCORDIUM_VERSION` should just be a parameter, or is the version of this plugin suppose to always match the version of `cargo-concordium`

limemloh:
> It looks like you could just parametrize this by the location of the binary for the relevant platform, and just have the one job build everything.
> 
> Or does the extension build step depend on the platform as well?

@abizjak 
We could, but I think it becomes a bit wonky to use then.
When packaging using `vsce`, we have to specify which target this is for, so we would need an additional variable for that.
When targeting windows, we need to append `.exe` in some places.
When targeting macos we need to use the signed executable

vscode-smart-contracts/scripts/linux-vscode-smart-contracts.Jenkinsfile
```diff
@@ -0,0 +1,72 @@
...
+        VERSION = sh(
+            returnStdout: true,
+            script: '''\
+                # Extract version number if not set as parameter
+                [ -z "$VERSION" ] && VERSION=$(node -p "require('vscode-smart-contracts/package.json').version")
+                echo -n "$VERSION"
+            '''.stripIndent()
+        )
+        CARGO_CONCORDIUM_VERSION = "${VERSION}"
+        OUTFILE = "s3://distribution.concordium.software/tools/linux/vscode-smart-contracts_${VERSION}"
```
Thahara, line 13: I'm assuming this is not just for Linux, so we could change that to vscode or plugins or something.

Thahara: nevermind, saw the other comments

limemloh: There will be a package for each platform as it contains the `cargo-concordium` binary.
VS code marketplace supports platform specific builds

vscode-smart-contracts/scripts/linux-vscode-smart-contracts.Jenkinsfile
limemloh: I don't think we want to have them match. I guess a more sane default would be to use the version in `cargo-concordium/Cargo.toml`.

vscode-smart-contracts/scripts/linux-vscode-smart-contracts.Jenkinsfile
```diff
@@ -0,0 +1,78 @@
+pipeline {
+    agent {
+        docker {
+            reuseNode true
+            image 'node:16-alpine'
```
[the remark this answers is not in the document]
limemloh: No probably not...

vscode-smart-contracts/scripts/linux-vscode-smart-contracts.Jenkinsfile
limemloh: I use `node:16` image instead

vscode-smart-contracts/scripts/linux-vscode-smart-contracts.Jenkinsfile
limemloh: I have changed it to use the version found in cargo.toml by default

abizjak:
> @abizjak We could, but I think it becomes a bit wonky to use then. When packaging using `vsce`, we have to specify which target this is for, so we would need an additional variable for that. When targeting windows, we need to append `.exe` in some places. When targeting macos we need to use the signed executable

Right. I was thinking that we would just supply the platform + version and the rest would be derived, something like this https://github.com/Concordium/concordium-node/blob/main/jenkinsfiles/ubuntu.Jenkinsfile#L10

abizjak (approved):

PR #58, opened 2023-04-28
Make vscode init project ensure cargo-generate is installed

limemloh:
## Purpose

Make vscode init project ensure cargo-generate is installed.

Also, fixes an issue where it would install it in the case of the prompt timeout.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/vscode-smart-contracts/src/commands.ts
+++ b/vscode-smart-contracts/src/commands.ts
@@ -15,20 +15,30 @@ const execFile = util.promisify(childProcess.execFile);
 /** The default name used for the output directory using the build command. */
 export const DEFAULT_OUT_DIR_NAME = "concordium-out";
 
+/** Wrap a function with try-catch displaying the error to the user, then throws the errors */
+export function displayErrorWrapper<A extends unknown[], B>(
+  fn: (...a: A) => B
+): (...a: A) => Promise<B> {
+  return async (...args: A) => {
+    try {
+      return await fn(...args);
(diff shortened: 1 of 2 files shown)


vscode-smart-contracts/src/commands.ts
```diff
@@ -222,3 +237,36 @@ export async function initProject() {
...
+async function haveCargoGenerateInstalled() {
+  const { stdout } = await execFile("cargo", ["--list"]);
+  console.log(stdout, lineStartingWithGenerate.test(stdout));
+  return lineStartingWithGenerate.test(stdout); // Check if the output have a line where the first word is `generate`.
+}
+
+/**
+ * Install the cargo-generate using cargo.
+ * The returned promise resolves when the install task has ended.
+ */
```
abizjak, line 255: Does this raise an exception if installation fails?

abizjak: How is that handled?

abizjak (approved):
If you can answer or resolve the issue of failing install this looks good to me.

vscode-smart-contracts/src/commands.ts
limemloh: It now returns the process exit code and fail if this is not 0.
I also ensured every failure is forwarded to the user, because VS Code will not show them by default.

abizjak: Great.

Issue #67, opened 2023-05-01
VSCode extension icon is not visible in dark mode


mh-concordium:
![image.png](https://images.zenhubusercontent.com/6284893f971a3daf4db6e569/0e359cdd-db1f-40eb-926a-a7e898df8578)

abizjak:
Closed by https://github.com/Concordium/concordium-smart-contract-tools/pull/71

PR #71, opened 2023-05-03
Update icon of the vscode extension

limemloh:
## Purpose

Update Icon of the VS Code extension to better support dark themes.
Closing #67 

Default light theme:
image
Default dark theme:
image


abizjak:
It would probably make sense to make the inner C bigger. Now it is quite tiny in the middle.

limemloh:
Here's the logo without text.
image
image

This seems better, and the Concordium name is visible everywhere anyway.

Bargsteen (approved):
Looks good!

abizjak:
> Here's the logo without text.
> 
> 
> 
> image
> 
> image
> 
> 
> 
> This seems better, and the Concordium name is visible everywhere anyway.

Yes I agree this is better. Let's go with this one.

abizjak (approved):
The one without text looks good.

PR #82, opened 2023-05-22
Implement `module_deploy_v1` for auto (re)compiling module to wasm

Bargsteen:
## Purpose

Implement `module_deploy_v1` for auto (re)compiling module to a deployable and testable Wasm module.

The method makes use of the synchronization primitive https://doc.rust-lang.org/std/sync/struct.Once.html to ensure that the module is only built once, instead of one time per test case. `Once` makes sure to block the other test threads until the module has been produced.

Example usage:
```
    module_build_v1().expect("Building succeeds");

    let module =
        module_load_v1("./concordium-test-out/module.wasm.v1").expect("Module exists and is valid");
```

## Changes

- Add a library part to cargo concordium `cargo_concordium_lib` to reuse the building functionality from it.
  - This needs to be published as well, which is not ideal but seems ok.
- Add the method `module_deploy_v1`

## Tasks remaining
- [ ] Use the library in the piggy bank tests after publishing
- [ ] Document how to use it in the developer documentation
   - [ ] Remove warnings about recompilation in the developer docs or at least refer to this method.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/cargo-concordium/Cargo.toml
+++ b/cargo-concordium/Cargo.toml
@@ -13,6 +13,14 @@ include = ["Cargo.toml", "Cargo.lock", "src/*"] # Only include the actual source
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[lib]
+name = "cargo_concordium_lib"
+path = "src/lib.rs"
+
+[[bin]]
+name = "cargo-concordium"
+path = "src/main.rs"
+
 [dependencies]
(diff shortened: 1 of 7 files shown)


abizjak (commented):
As we discussed in the office this is a bit of a complex solution since it brings in another dependency which is primarily a binary package.

We will instead attempt to expand cargo-concordium to run integration tests by expanding `cargo concordium test`.

PR #84, opened 2023-05-23
Compile module and run integration tests when using `cargo concordium…

Bargsteen:
## Purpose

Extend `cargo concordium test` such that it compiles the module and runs all integration tests.

## Changes

- Make changes to the test handling.

## TODOs

- [ ] Should it output in `concordium-test-out` or `concordium-out` (also used by VSCode extension).
- [ ] Should we allow the user to specify
   - Where to output the module compiled
   - Schema options for building
   - Extra cargo args
   - Or will the above be handled by the user manually using `cargo concordium build`?
- [ ] Should it compile the module even when there are no integration tests?

## Checklist

      hard-to-understand areas.

### DIFF
--- a/cargo-concordium/src/build.rs
+++ b/cargo-concordium/src/build.rs
@@ -33,6 +33,7 @@ use std::{
 /// Encode all base64 strings using the standard alphabet and padding.
 const ENCODER: base64::engine::GeneralPurpose = general_purpose::STANDARD;
 
+/// Convert a string to snake case by replacing `-` with `_`.
 fn to_snake_case(string: &str) -> String { string.to_lowercase().replace('-', "_") }
 
 #[derive(Debug, Clone, Copy)]
@@ -659,6 +660,71 @@ pub fn write_json_schema_to_file_v3(
     write_schema_json(path_of_out, contract_name, contract_counter, schema_json)
 }
 
+/// Build the smart contract module and run all integration tests.
+///
+/// The module is built with [`build_contract`] and is placed at
+/// `./concordium-test-out/module.wasm.v1`.
+///
+/// The method discovers all test targets and run them.
+pub fn build_and_run_integration_tests() -> anyhow::Result<()> {
+    eprintln!(
+        "\n{}",
+        Color::Green.bold().paint("Running integration tests ...")
+    );
+
+    let metadata = MetadataCommand::new()
+        .exec()
+        .context("Could not access cargo metadata.")?;
+
+    let mut cargo_args = vec!["test"];
+    let test_args = metadata
+        .root_package()
+        .context("Could not determine package.")?
+        .targets
+        .iter()
+        .filter_map(|t| {
+            if t.kind == ["test"] {
+                Some(["--test", &t.name])
+            } else {
+                None
+            }
+        })
+        .flatten();
+    cargo_args.extend(test_args);
+
+    // TODO: Add option for user to embed schema (and also generate the json
+    // schema)?
+    build_contract(
+        WasmVersion::V1,
+        SchemaBuildOptions::DoNotBuild,
+        Some("concordium-test-out/module.wasm.v1".into()),
+        &[],
+    )?;
+
+    // Output what we are doing so that it is easier to debug if the user
+    // has their own features or options.
+    eprintln!(
+        "{} cargo {}",
+        Color::Green.bold().paint("Running"),
+        cargo_args.join(" "),
+    );
+
+    let result = Command::new("cargo")
+        .args(cargo_args)
+        .stdout(Stdio::inherit())
+        .stderr(Stdio::inherit())
+        .output()
+        .context("Failed running integration tests.")?;
+
+    anyhow::ensure!(
+        result.status.success(),
+        Color::Red
+            .bold()
+            .paint("One or more integration tests failed.")
+    );
+    Ok(())
+}
+
 /// Build tests and run them. If errors occur in building the tests, or there
 /// are runtime exceptions that are not expected then this function returns
 /// Err(...).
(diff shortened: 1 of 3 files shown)


cargo-concordium/src/build.rs
```diff
@@ -33,6 +33,7 @@ use std::{
 /// Encode all base64 strings using the standard alphabet and padding.
 const ENCODER: base64::engine::GeneralPurpose = general_purpose::STANDARD;
 
+/// Convert a string to snake case by replacing `-` with `_`.
```
abizjak, line 36: Maybe the functionality is clear from the code, but not the purpose.

Maybe add that this is what rust does as well when converting crate names (which often have `-`) to module names (which cannot have that).

cargo-concordium/src/build.rs
```diff
@@ -659,6 +660,71 @@ pub fn write_json_schema_to_file_v3(
...
+    cargo_args.extend(test_args);
+
+    // TODO: Add option for user to embed schema (and also generate the json
+    // schema)?
+    build_contract(
+        WasmVersion::V1,
+        SchemaBuildOptions::DoNotBuild,
+        Some("concordium-test-out/module.wasm.v1".into()),
+        &[],
+    )?;
```
abizjak, line 702: I think the reasonable behaviour would be the following
- support the same options as `build` does.
- if the user does not supply the output path then build it in the usual location in `target/...` (this is also the behaviour of build)

This way we don't pollute the user's root directory unless they tell us so.

limemloh (commented):
My take on the questions in the PR:
- I would expect it to support all the builds options, since user might want to use the produced module, otherwise this command is not really helping much.
- I think it should compile the module every time, we could provide a flag for only running the unit tests in wasm, where it could skip compiling the module.

cargo-concordium/src/main.rs
```diff
@@ -187,61 +196,84 @@ enum Command {
...
-            raw = true,
-            help = "Extra arguments passed to `cargo build` when building Wasm module."
-        )]
-        cargo_args:        Vec<String>,
+        #[structopt(flatten)]
+        build_options: BuildOptions,
     },
 }
 
+// TODO: A doc comment here replaces the `about` string in Test and Build.
```
abizjak, line 204: Can't we add an explicit `about` where it is used then?

What other options should we consider here?

cargo-concordium/src/main.rs
```diff
@@ -442,10 +471,25 @@ pub fn main() -> anyhow::Result<()> {
...
-        Command::Test { args, seed } => {
-            let success =
-                build_and_run_wasm_test(&args, seed).context("Could not build and run tests.")?;
-            ensure!(success, "Test failed");
+        Command::Test {
+            seed,
+            build_options,
+            only_unit_tests,
+        } => {
+            // TODO: Fail early if wasm version == 0?
```
abizjak, line 479: I think we don't need to do that. Maybe they have written some tests that work with V0 modules in some other way.

Does that make sense?

cargo-concordium/src/main.rs
```diff
@@ -58,15 +59,23 @@ enum Command {
...
+        #[structopt(flatten)]
+        build_options:   BuildOptions,
         #[structopt(
-            raw = true,
-            help = "Extra arguments passed to `cargo build` when building the test Wasm module."
+            name = "only-unit-tests",
+            long = "only-unit-tests",
+            help = "Whether only the unit tests should be run."
         )]
-        args: Vec<String>,
```
abizjak, line 69: I think we should keep this argument unless it causes problems with the `raw` argument of `BuildOptions`.

The normal `cargo test` allows you to say `cargo test foo` and it'll only run the tests that matches `foo` (not sure what "matches" is exactly, but is is somethlng like substring match).

If we kept this we could also do this. instead of adding all `--test` flags we would only add it to the tests that match.

abizjak: If just positional arguments don't work maybe we can use named arguments instead. The API won't be the same as cargo test, but I think that's acceptable if necessary.

abizjak (commented):
Looks good. 

I only have two main concerns related to the command-line API that we should resolve.

cargo-concordium/src/main.rs
Bargsteen: Yes, that is reasonable. I'll remove the TODO.

cargo-concordium/src/main.rs
Bargsteen: There is already an explicit `about`, but this seems to take precedence. I'll look into other options.

Bargsteen: Without doc comments on `BuildOptions`:
image

With doc comments:
image

abizjak: I see. This macro is a bit too magical :) 

I guess we should then just have a comment to make sure somebody does not add a doc comment in the future :)

cargo-concordium/src/main.rs
Bargsteen: It is included in the `BuildOptions`, so it is still there. However, we could consider adding one set of args for `cargo build` and one set for `cargo test`. Right now the extra args are used for both, which might pose a problem.

Bargsteen: As discussed on slack, it is not possible to have two `raw` args, i.e. ones that capture all arguments after `cargo concordium test -- `, so we went for an option where the user can limit which test targets to run with the `--test` flag.

cargo-concordium/src/main.rs
Bargsteen: Yep 🪄🐇

cargo-concordium/src/main.rs
Bargsteen: Note that it is only possible to specify test _targets_ with `--test`. It won't try to match on the individual test names. So you need to split your tests into multiple files if you want the flexibility to run a subset of them.

abizjak: That's reasonable.

limemloh (approved):
LGTM

PR #87, opened 2023-06-13
Deploying/Initialization of smart contracts via a front end

### LINKED ISSUE #88: Building a front end that smart contract developers can use to deploy/initialize their smart contracts
**Task description**

To simplify the deployment process, we want to provide a front end that developers can use. That way new smart contract developers (e.g. at a hackathon) get easily started without downloading/setting up `concordium-client` .

### LINKED ISSUE #92: front-end-tool: Check for duplicate module
**Description**

It would be useful if the front-end checked if the module the user is trying to deploy is already deployed on the relevant network instead of letting the transaction through, and it failing.

DOBEN:
## Purpose

closes #88 
closes #92 

Front end to deploy/initialize smart contracts using the browser wallet.

## Changes

- Add deploy flow
- Add initialization flow

## Checklist

      hard-to-understand areas.


front-end-tools/Jenkinsfile
```diff
@@ -0,0 +1,35 @@
...
+        }
+        stage('build-push') {
+            steps {
+                sh '''\
+                    docker build \
+                      --build-arg build_image="${build_image}" \
+                      --label build_image="${build_image}" \
+                      --tag="${image_name}" \
+                      -f ./front-end-tools/Dockerfile \
+                      .
```
limemloh, line 29: Unless we need something outside of the `front-end-tools` directory, it would be better to set the build context to this directory. Since the context is now the root of the repo docker will copy every file in the repo before it starts running the dockerfile, which will become slower as this repo grows (and the more local files you have in the repo)

front-end-tools/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+FROM ${build_image} AS build
+
+RUN yarn set version 3.2.0
+
+# Copy front end files
+WORKDIR /app
+COPY ./front-end-tools ./
+
+WORKDIR /app/front-end-tools
+RUN yarn install && yarn build
```
limemloh, line 14: We should have a `.dockerignore` file and have docker ignore `node_modules` otherwise your local build of packages might end in the image. Will also speed up the docker build time.

front-end-tools/README.md
```diff
@@ -0,0 +1,81 @@
...
+- Upload and deploy a smart contract module to chain.
+- Initialize a smart contract module on chain.
+
+Only the browser wallet (no walletConnect) is supported in the first version.
+
+## Prerequisites
+
+-   Browser wallet extension must be installed in Chrome browser.
+
+## Running the front end
```
limemloh, line 14: This section and below is meant for people developing this, right?
Should we have a section for how to use the docker image that we have pushed to docker hub?

limemloh (changes requested):
Seems to be missing the `package.json`, unable to build it without

DOBEN:
> Seems to be missing the `package.json`, unable to build it without

Right, in the `.gitignore` were all files excluded that match the pattern `*.json` (The rule came originally from our Rust repos where `*.json` files are usually inputParameterJSON files used for testing locally so we exclude them from `git`). I removed that `*.json` in `.gitignore` and the `package.json` is now committed to git.

front-end-tools/src/constants.ts
```diff
@@ -0,0 +1,16 @@
...
+import moment from 'moment';
+
+export const REFRESH_INTERVAL = moment.duration(10, 'seconds');
+
+// The TESTNET_GENESIS_BLOCK_HASH can be used to check that the user has its browser wallet connected to testnet.
+export const TESTNET_GENESIS_BLOCK_HASH = '4221332d34e1694168c2a0c0b3fd0f273809612cb13d000d5c2e00e85f50f796';
+export const TESTNET: Network = {
+    name: 'testnet',
+    genesisHash: TESTNET_GENESIS_BLOCK_HASH,
+    jsonRpcUrl: 'https://json-rpc.testnet.concordium.com',
```
abizjak, line 11: Why do we still keep using the jsonRpc client?

This is deprecated now, so it would be better not to rely on it in new projects.

front-end-tools/src/reading_from_blockchain.ts
```diff
@@ -0,0 +1,5 @@
+import { JsonRpcClient } from '@concordium/web-sdk';
+
+export async function accountInfo(rpcClient: JsonRpcClient, account: string) {
```
abizjak, line 3: COuld we please not use the jsonrpcclient seeing it's deprecated?

is there something preventing that?

front-end-tools/Jenkinsfile
DOBEN: updated

front-end-tools/Dockerfile
DOBEN: Instead of `.dockerignore`, I specified the 5 files that should be copied into the docker image.

front-end-tools/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "packageManager": "yarn@3.2.0",
+    "version": "1.0.0",
+    "license": "Apache-2.0",
+    "engines": {
+        "node": ">=16.x"
+    },
+    "dependencies": {
+        "@concordium/browser-wallet-api-helpers": "^2.5.0",
+        "@concordium/react-components": "^0.3.0",
+        "@concordium/wallet-connectors": "^0.3.1",
```
bisgardo, line 12: All symbols from `wallet-connectors` are re-exposed in `react-components` so there's no need to explicitly depend on it (just make sure that you're imports refer to `react-components` only).

front-end-tools/src/constants.ts
```diff
@@ -0,0 +1,18 @@
+import { BrowserWalletConnector, ephemeralConnectorType, Network } from '@concordium/react-components';
+import moment from 'moment';
+
+export const REFRESH_INTERVAL = moment.duration(10, 'seconds');
+
+// The TESTNET_GENESIS_BLOCK_HASH can be used to check that the user has its browser wallet connected to testnet.
+export const TESTNET_GENESIS_BLOCK_HASH = '4221332d34e1694168c2a0c0b3fd0f273809612cb13d000d5c2e00e85f50f796';
+export const TESTNET: Network = {
```
bisgardo, line 8: This constant is now defined in the dapp-libs, so you can import it instead of defining it.

bisgardo (commented):
I suggest bootstrapping project like this with [CRA](https://create-react-app.dev/) or [vite](https://vitejs.dev/guide/). That provides sane default config and very useful tooling like hot reload out of the box with zero effort. I thinks that's better than continuously forking a lot of complex configuration that at least I don't really understand. You can still customize, but I haven't found the need to.

See for example the [contractupdate](https://github.com/Concordium/concordium-dapp-libraries/tree/main/samples/contractupdate) sample dapp which uses CRA.

front-end-tools/README.md
```diff
@@ -0,0 +1,81 @@
+# Developer Front End Tools
+
+The front end contains useful tools for developers:
```
Bargsteen, line 3:
```suggestion
The front end contains useful functionalities for smart contract developers:
```

front-end-tools/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+- Upload and deploy a smart contract module to chain.
+- Initialize a smart contract module on chain.
+
+Only the browser wallet (no walletConnect) is supported in the first version.
```
Bargsteen, line 8:
```suggestion
Only the browser wallet (not walletConnect) is supported in the first version.
```

Bargsteen (approved):
I tried it out and it works very well. Nicely done!

DOBEN:
```
I suggest bootstrapping a project like this with [CRA](https://create-react-app.dev/) or [vite](https://vitejs.dev/guide/) ...
```

I agree that for the next front-end project, it would be a good exercise to start with CRA (since it is a very common framework and probably easier understandable by outside people than our current front-end setup). For this front-end now, I think it is not worth the effort to re-write it in CRA and we already have several front-end with a similar setup internally so internally we are used to that setup.

Bargsteen (approved):

front-end-tools/src/Main.tsx
```diff
@@ -0,0 +1,579 @@
...
+                                    }ccdscan.io/?dcount=1&dentity=account&daddress=${account}`}
+                                    target="_blank"
+                                    rel="noreferrer"
+                                >
+                                    {account}
+                                </a>
+                            </div>
+                            <br />
+                            <div className="label">Your account balance:</div>
+                            <div>{accountBalance.replace(/(\d)(?=(\d\d\d\d\d\d)+(?!\d))/g, '$1.')} CCD</div>
```
abizjak, line 245: Screenshot from 2023-06-22 09-01-05

It would be ideal if I was not allowed to try to deploy the module if I don't have enough balance. But there might not be good helpers in the web SDK to allow you to precompute this easily.

Also, this looks like a hacky way of displaying CCD amounts with a `.`. Is there no library function in the Web SDK to display amounts in CCD? If there isn't it should be added.

abizjak (changes requested):
it's a good start of a tool.

I have some mainly UI/UX comments, some of which I think should be done before we deploy it, and some we should make tasks for for later.

front-end-tools/src/Main.tsx
```diff
@@ -0,0 +1,579 @@
...
+                                </label>
+                                <br />
+                                <label className="field">
+                                    Add Smart Contract Name:
+                                    <br />
+                                    <input
+                                        className="inputFieldStyle"
+                                        id="initName"
+                                        type="text"
+                                        placeholder="myContract"
```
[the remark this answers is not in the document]
DOBEN: Let's address comments that rely on support/parsing of the "embedded schemas" from the module in version 2.0 of this front end.
There will be better support for this soon available:
`https://github.com/Concordium/concordium-contracts-common/issues/96`
 
 I see you already created a task for that, so I will close this thread.
https://github.com/Concordium/concordium-smart-contract-tools/issues/93

front-end-tools/src/Main.tsx
```diff
@@ -0,0 +1,579 @@
...
+                                <label className="field">
+                                    Add Module Reference:
+                                    <br />
+                                    <input
+                                        className="inputFieldStyle"
+                                        id="moduleReference"
+                                        type="text"
+                                        placeholder="91225f9538ac2903466cc4ab07b6eb607a2cd349549f357dfdf4e6042dde0693"
+                                        onChange={changeModuleReferenceHandler}
+                                    />
```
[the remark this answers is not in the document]
DOBEN: Users can now copy the module reference from the front end.

front-end-tools/src/Main.tsx
```diff
@@ -0,0 +1,579 @@
...
+                            </TestBox>
+                            <TestBox header="Step 2: Initialize Smart Contract">
+                                <label className="field">
+                                    Add Module Reference:
+                                    <br />
+                                    <input
+                                        className="inputFieldStyle"
+                                        id="moduleReference"
+                                        type="text"
+                                        placeholder="91225f9538ac2903466cc4ab07b6eb607a2cd349549f357dfdf4e6042dde0693"
```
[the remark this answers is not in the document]
DOBEN: It is a valid reference but it is just a `placeholder`, meaning the `value` of this field is not set to `91225...`, so I don't see a thread where people accidentally deploy module `91225...`. I leave it as it is if ok.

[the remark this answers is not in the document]
abizjak: Ok

front-end-tools/src/Main.tsx
DOBEN: https://github.com/Concordium/concordium-node-sdk-js/issues/215
It would be good to have a simulate transaction/feature in the webSDK, so a front-end can first call to simulate a specific transaction to see if it would revert (e.g. because insufficient funds in the wallet or because of a contract revert) and only if the simulation was successful, allow the user the execution of the real transaction on the front end. It would be similar to the invokeContract call (which can simulate Transaction.Update transactions). This request is to expand the simulation to any type of transaction e.g. deploy/initilize transactions as well.

This display of CCD is something that should be done in the browser, e.g. get the browser setting and based on your language/region, you either use a '.' or a ',' after the full CCD amount. I feel that is not something the Web SDK can do. 

What the web SDK could do is convert `1 CCD` into 1000000 microCCD which would be nice.

front-end-tools/src/Main.tsx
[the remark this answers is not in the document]
DOBEN: The front end has now a `checkBox` that you can click `Use Module Reference from Step 1`.

It will then autofill the `moduleReference` form Step 1 to Step2. In detail, it will first use `moduleReferenceDeployed` (meaning the `moduleReference` that you just deployed to chain in Step 1 (or if that value is not available) the `moduleReferenceCalculated` if you just uploaded a module file in Step 1 but did not deploy it to chain.

front-end-tools/src/Main.tsx
abizjak: As we discussed in the planning meeting, this needs some node support first, which we will add in Q3. After that is doen we should definitely have this.

front-end-tools/src/writing_to_blockchain.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    return connection.signAndSendTransaction(
+        account,
+        AccountTransactionType.InitContract,
+        {
+            amount: new CcdAmount(BigInt(amount ? Number(amount) : 0)),
+            moduleRef: new ModuleReference(moduleReference),
+            initName,
+            param: toBuffer(''),
+            maxContractExecutionEnergy: 30000n,
```
abizjak, line 89: Shouldn't this be configurable as well? It might not be enough, or it might be too expensive.

Is it very hard to add a field with this default for initializing?

abizjak (approved):
Good improvements on the first version. Looks good and smoother now.

While testing I found the following error

Screenshot from 2023-06-27 12-44-46

The "account not found" has quoted spaces in it `%20%20`. Can you fix that?

Issue #90, opened 2023-06-20
Get chain parameters for testing library via node connection


Bargsteen:
**Task description**

We want to expand the testing library with various ways of using data from an actual chain via a node.

The first step on that journey is to allow users to get certain chain parameters from the chain, including the exchange rates and block time.

The Concordium Rust SDK already supports communicating with a node, so we will add it as a dependency to the testing library.
The calls to a node take time and are `async` in the Rust SDK, but for testing, we want to hide the asynchronicity from the user by blocking the thread until the task is finished.

The reason for doing so is that it really isn't possible to run any parts of an individual test in parallel.
Hiding the async functions also provide a simpler and cleaner interface for the users.

**Sub-tasks**
- [ ] Add a constructor function to `Chain` which connects to a running node.
- [ ] Add a method for setting the `Chain`s exchange rate parameters and the block time to match those of the connected node.

PR #91, opened 2023-06-21
Add methods for setting exchange rates via external node

Bargsteen:
## Purpose

Closes https://github.com/Concordium/concordium-smart-contract-tools/issues/90

## Changes

- Add methods to the `Chain`
   - `block_time` and `tick_block_time` for getting and increasing the block time.
   - `external_query_block` for getting the default block used for external queries.
 - Make the `block_time` field `pub(crate)` instead of `pub` since it is not possible to get the `ChainParameters` anyhow.
- Add the Rust SDK and Tokio as dependencies
- Add some error types
- Re-export the following types:
   - `BlockHash`
   - `Timestamp` since it makes more sense to use than the alias `SlotTime` now
   - `Duration`
   - `Endpoint`
- Add a builder pattern to the `Chain` with `ChainBuilder`.
   - Supports setting up an external node connection
   - Supports setting exchange rates and the block time explicitly or via the external node
- Add tests for the `ChainBuilder`
- The tests that use I/O are in the `io_tests` module and they are skipped by the CI due to the `-- --skip io_tests` flag that I added in there.

## Checklist

      hard-to-understand areas.


limemloh (approved):
LGTM

concordium-smart-contract-testing/src/types.rs
```diff
@@ -600,3 +632,53 @@ pub struct ExchangeRateError;
 #[derive(Debug, Error)]
 #[error("Any signer must have at least one key.")]
 pub struct ZeroKeysError;
+
+/// Errors that occur while setting up the connection to an external node.
+#[derive(Debug, thiserror::Error, PartialEq, Eq)]
+pub enum SetupExternalNodeError {
+    /// It was not possible to connect to a node on the provided endpoint.
+    #[error("Could not connect to the provided endpoint.")]
+    CannotConnect,
```
abizjak, line 641: Why does this not have an inner error that would tell why the connection could not be established?

concordium-smart-contract-testing/src/types.rs
```diff
@@ -600,3 +632,53 @@ pub struct ExchangeRateError;
...
+
+/// Errors that occur while trying to communicate with an external node.
+#[derive(Debug, Error, PartialEq, Eq)]
+pub enum ExternalNodeError {
+    /// An external node has not been configured.
+    #[error("An external node has not been configured.")]
+    NotConfigured,
+    /// The query could not be performed.
+    #[error("Could not perform the query.")]
+    QueryError,
```
abizjak, line 658: Why does this not have any more details as to why it failed?

concordium-smart-contract-testing/src/types.rs
```diff
@@ -35,22 +37,52 @@ pub(crate) struct ChainParameters {
...
-    pub(crate) contracts:           BTreeMap<ContractAddress, Contract>,
+    pub(crate) contracts:                BTreeMap<ContractAddress, Contract>,
     /// Next contract index to use when creating a new instance.
-    pub(crate) next_contract_index: u64,
+    pub(crate) next_contract_index:      u64,
+    /// An optional connection to an external node.
+    pub(crate) external_node_connection: Option<ExternalNodeConnection>,
+}
+
+/// A builder for the [`Chain`].
```
abizjak, line 75: Can we add a pointer on how to get a chain out of it, a link to a `.build` function.

concordium-smart-contract-testing/Cargo.toml
```diff
@@ -16,8 +16,10 @@ exclude = ["tests"] # Do not publish tests.
 concordium_base = {version = "2.0", path = "../concordium-base/rust-src/concordium_base"}
 concordium-smart-contract-engine = {version = "2.0", path = "../concordium-base/smart-contracts/wasm-chain-integration"}
 concordium-wasm = {version = "2.0", path = "../concordium-base/smart-contracts/wasm-transform"}
+concordium-rust-sdk = "2.4"
```
abizjak, line 19: I guess this will be version 3 instead.

concordium-smart-contract-testing/src/impls.rs
```diff
@@ -81,33 +90,188 @@ impl ChainParameters {
...
+    /// # Example
+    ///
+    /// ```no_run
+    /// # use concordium_smart_contract_testing::*;
+    /// let chain = Chain::builder()
+    ///     .external_node_connection("http://node.testnet.concordium.com:20000")
+    ///     .build()
+    ///     .unwrap();
+    /// ```
+    pub fn external_node_connection(mut self, endpoint: &str) -> Self {
```
abizjak, line 133: Why is this not an `Endpoint` or `Into<Endpoint>`? That is what we always use in Rust SDK examples.

I don't like using strings. In particular it means this library now has to have "malformed URL" error, which it does not need to have.

concordium-smart-contract-testing/src/impls.rs
```diff
@@ -1056,11 +1223,277 @@ impl Chain {
...
+        let timestamp_before_or_after =
+            connection.with_client(block, |block_identifier, mut client| async move {
+                let response = client.get_block_info(block_identifier).await?;
+                let block_time = response.response.block_slot_time.timestamp_millis();
+                Ok((response.block_hash, block_time))
+            })?;
+
+        // Our `Timestamp` only supports times *after* the unix epoch, so we use `0` if
+        // the queried timestamp is negative.
+        let timestamp_after: u64 = i64::max(timestamp_before_or_after, 0) as u64;
```
abizjak, line 1365: Just use as u64, the node does not support timestamps before unix epoch anywhere.

concordium-smart-contract-testing/src/impls.rs
```diff
@@ -1056,11 +1223,277 @@ impl Chain {
...
+            runtime,
+            sticky_block: None,
+        });
+
+        Ok(())
+    }
+
+    /// Try to get the [`ExternalNodeConnection`] or return an error.
+    ///
+    /// The connection is only available, if the [`Chain`] has been created with
```
abizjak, line 1417: Something is missing.

concordium-smart-contract-testing/src/impls.rs
```diff
@@ -1056,11 +1223,277 @@ impl Chain {
...
     pub fn euro_per_energy(&self) -> ExchangeRate { self.parameters.euro_per_energy }
+
+    /// Return the current block time.
+    pub fn block_time(&self) -> Timestamp { self.parameters.block_time }
+
+    /// Return the "sticky block", i.e. the block that is saved and will be used
+    /// by default for future queries.
+    ///
+    /// See [`set_block_time_via_external_node`][Chain::set_block_time_via_external_node] for a more detailed explanation.
+    pub fn sticky_block(&self) -> Option<BlockHash> {
```
abizjak, line 1440: I find it a bit odd to call this "sticky". I'm trying to think if I could guess as to what it is if I did not know, but I cannot imagine doing so.

Have you consider something more like `query_block`?

concordium-smart-contract-testing/src/impls.rs
```diff
@@ -1056,11 +1223,277 @@ impl Chain {
...
+            timeout(EXTERNAL_NODE_TIMEOUT_DURATION, f(block_identifier, client))
+                .await
+                .map_err(|_| ExternalNodeError::Timeout)?
+        });
+        // Extract the return value and maybe set the stick block.
+        match res {
+            Ok((block_hash, return_value)) => {
+                if self.sticky_block.is_none() {
+                    self.sticky_block = Some(block_hash)
+                }
```
abizjak, line 1491: I would probably expect more the behaviour that "sticky" block is the last block that was used.

So it is overridden even if it is set already, and I supply a custom block.

concordium-smart-contract-testing/src/impls.rs
```diff
@@ -1301,6 +1734,22 @@ pub(crate) fn contract_events_from_logs(logs: v0::Logs) -> Vec<ContractEvent> {
...
+    fn from(_: concordium_rust_sdk::endpoints::Error) -> Self { Self::CannotConnect }
+}
+
+impl From<ExternalNodeNotConfigured> for ExternalNodeError {
+    fn from(_: ExternalNodeNotConfigured) -> Self { Self::NotConfigured }
+}
+
+impl From<ExchangeRateError> for ChainBuilderError {
+    fn from(_: ExchangeRateError) -> Self { Self::ExchangeRateError }
+}
```
abizjak, line 1751: Why are we forgetting the inner error?

I think those generally provide useful information that can be used to debug what went wrong.

concordium-smart-contract-testing/src/impls.rs
```diff
@@ -1378,4 +1827,169 @@ mod tests {
...
+    fn test_block_time_from_unknown_block() {
+        let mut chain = Chain::builder()
+            .external_node_connection("http://node.testnet.concordium.com:20000")
+            .build()
+            .unwrap();
+        let err = chain
+            .set_block_time_via_external_node(Some(
+                BlockHash::from_str(
+                    "4f38c7e63645c59e9bf32f7ca837a029810b21c439f7492c3cebe229a2e3ea07", // A block from mainnet.
+                )
```
abizjak, line 1989: Please use `parse` for better readability.

concordium-smart-contract-testing/src/impls.rs
```diff
@@ -1378,4 +1827,169 @@ mod tests {
...
+        chain.set_block_time_via_external_node(None).unwrap();
+        let sticky_block_2 = chain.sticky_block().unwrap();
+        let block_time_2 = chain.block_time();
+        // Check that the sticky block hasn't changed.
+        assert_eq!(sticky_block_1, sticky_block_2);
+        // Check that the block matches the one specified initially.
+        assert_eq!(sticky_block_1,
+                BlockHash::from_str(
+                    "95ff82f26892a2327c3e7ac582224a54d75c367341fbff209bce552d81349eb0",
+                )
```
abizjak, line 1968: Can we please use `parse` so it's a bit less verbose?

concordium-smart-contract-testing/src/impls.rs
```diff
@@ -1378,4 +1827,169 @@ mod tests {
...
+        let block_time_1 = chain.block_time();
+        let sticky_block_1 = chain.sticky_block().unwrap();
+
+        // Set block time using a specific block that exists.
+        chain
+            .set_block_time_via_external_node(Some(
+                BlockHash::from_str(
+                    "95ff82f26892a2327c3e7ac582224a54d75c367341fbff209bce552d81349eb0",
+                )
+                .unwrap(),
```
abizjak, line 1930:
```suggestion
                "95ff82f26892a2327c3e7ac582224a54d75c367341fbff209bce552d81349eb0".parse().unwrap()
```

abizjak (commented):
Looks OK overall.

I'm not very happy with hiding inner errors since it makes it harder to debug.

I would also simplify the "sticky" block to make it always the block that was used in the latest query, or LastFinal if there was no such query.

.github/workflows/ci.yaml
```diff
@@ -159,4 +159,4 @@ jobs:
           override: true
       - name: Test
         working-directory: ${{ matrix.build-dir }}
-        run: cargo test
+        run: cargo test -- --skip io_tests # Skip the I/O tests in the testing library.
```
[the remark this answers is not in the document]
Bargsteen: Yes.

concordium-smart-contract-testing/src/types.rs
Bargsteen: Added now.

concordium-smart-contract-testing/src/types.rs
Bargsteen: Yes.

concordium-smart-contract-testing/src/impls.rs
Bargsteen: Good point.

concordium-smart-contract-testing/src/impls.rs
Bargsteen: Ok, nice.

concordium-smart-contract-testing/src/impls.rs
Bargsteen: Yes, added now.

concordium-smart-contract-testing/src/impls.rs
Bargsteen: Changed now.

concordium-smart-contract-testing/src/impls.rs
```diff
@@ -81,33 +87,339 @@ impl ChainParameters {
...
+            chain.parameters.euro_per_energy = euro_per_energy;
+        }
+
+        // Check the exchange rates and return early if they are invalid.
+        check_exchange_rates(
+            chain.parameters.euro_per_energy,
+            chain.parameters.micro_ccd_per_euro,
+        )?;
+
+        match (self.block_time, self.block_time_from_external) {
```
abizjak, line 363: Couldn't we prevent this already in the builder while calling these methods? And so we would not have to fail here in `build`?

abizjak (approved):

concordium-smart-contract-testing/src/impls.rs
Bargsteen: We could, but then we would have to return a `Result` in some of the configuration methods (but not all), which I don't think is ideal. As far as I could tell, the typical way is to only have the build method return a `Result`.
Or do you have another way of doing it in mind?

concordium-smart-contract-testing/src/types.rs
```diff
@@ -600,3 +641,116 @@ pub struct ExchangeRateError;
...
+pub enum SetupExternalNodeError {
+    /// It was not possible to connect to a node on the provided endpoint.
+    #[error("Could not connect to the provided endpoint due to: {error}")]
+    CannotConnect {
+        /// The inner error.
+        #[from]
+        error: sdk::endpoints::Error,
+    },
+    /// Request timed out.
+    #[error("The request timed out.")]
```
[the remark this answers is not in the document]
Bargsteen: Well, it is used both for creating the Client and for querying the last final block or the external_query_block.
Both time out after 10 seconds.
I can change it to "connection" if you think that is more fitting, or we can have separate cases.

concordium-smart-contract-testing/Cargo.toml
Bargsteen, line 19:
```suggestion
concordium-rust-sdk = "3.0"
```

concordium-smart-contract-testing/src/impls.rs
abizjak: No, that would be the result. I thought using `?` would make it pretty painless.

But I'm ok the way it is if you think it's better. I guess in practice the error will be in the "same place", when they run.

concordium-smart-contract-testing/src/types.rs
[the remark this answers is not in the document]
abizjak: They are in principle different timeouts so it would be cleanest to have two constants.

The initial connect timeout should usually be shorter.

concordium-smart-contract-testing/src/impls.rs
Bargsteen: Ok, then let's keep it as is.

concordium-smart-contract-testing/src/types.rs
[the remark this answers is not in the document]
Bargsteen: I've added a separate case for this now and made some clarifications. I've set the timeout duration for the initial connection to 3 seconds.

concordium-smart-contract-testing/src/types.rs
```diff
@@ -600,3 +641,119 @@ pub struct ExchangeRateError;
...
+#[derive(Debug, Error, PartialEq, Eq)]
+#[error("An external node has not been configured.")]
+pub struct ExternalNodeNotConfigured;
+
+#[derive(Debug, Error)]
+pub enum ChainBuilderError {
+    /// The provided exchange rates are not valid.
+    /// Meaning that they do not correspond to one energy costing less than
+    /// `u64::MAX / `
+    /// [`concordium_base::constants::MAX_ALLOWED_INVOKE_ENERGY`]`.
```
limemloh, line 704: This constant is no longer in base

limemloh (approved):

concordium-smart-contract-testing/src/types.rs
Bargsteen: Hmm, you are right. I see that Ales removed it again. Has it been moved elsewhere, or why was it removed @abizjak ? 😊

abizjak: It was removed because the node API has no such concept anymore and it does not even require you to supply `max_energy`. 

Instead the node itself has a configuration option to prevent DOS and it will decide the maximum energy allowed.

YOu can still supply an NRG amount if you want to test something, but the NRG amount that will actually be used is min(supplied amount, node configuration option).

Bargsteen: As discussed with Ales, I've replaced the constant with the value instead,

concordium-smart-contract-testing/Cargo.toml
```diff
@@ -7,20 +7,23 @@ license = "MPL-2.0"
...
+concordium-smart-contract-engine = {version = "3.0", path = "../concordium-rust-sdk/concordium-base/smart-contracts/wasm-chain-integration"}
+concordium-wasm = {version = "3.0", path = "../concordium-rust-sdk/concordium-base/smart-contracts/wasm-transform"}
+concordium-rust-sdk = {version = "3.0", path = "../concordium-rust-sdk"}
+tokio = { version = "1.28", features = ["rt-multi-thread", "time"] }
 sha2 = "0.10"
 anyhow = "1"
 thiserror = "1.0"
 num-bigint = "0.4"
 num-integer = "0.1"
+toml_edit = "=0.19.14" # Fix the version to .14, as .15 has MSRV 1.66. Remove this line when upgrading out MSRV.
```
abizjak, line 26: Since we bumped MSRV in concordium-std to 1.66 we might as well do it here now.

concordium-smart-contract-testing/src/impls.rs
```diff
@@ -1065,11 +1387,245 @@ impl Chain {
...
+impl ExternalNodeConnection {
+    /// Execute an async task with the [`sdk::v2::Client`].
+    ///
+    /// If a block is provided, it will be used for the query. Otherwise, it
+    /// will use the default query block.
+    ///
+    /// If the task takes longer than [`EXTERNAL_NODE_TIMEOUT_DURATION`] then
+    /// the connection times out and an [`Err(ExternalNodeError::Timeout)`] is
+    /// returned.
+    fn with_client<T, F, Fut>(
```
abizjak, line 1605: I think it might be useful to clarify here taht this function cannot be nested, that'll lead to a panic.

abizjak (approved):

concordium-smart-contract-testing/src/impls.rs
```diff
@@ -1065,11 +1387,245 @@ impl Chain {
...
+        &mut self,
```
abizjak, line 1606: This can take a `&self`

Bargsteen: Ah, nicely spotted ;)

concordium-smart-contract-testing/src/impls.rs
```diff
@@ -1296,7 +1854,8 @@ pub(crate) fn lookup_module_cost(module: &ContractModule) -> Energy {
 /// ```
 ///
 /// To convert the `energy` parameter to mCCD (the vertical lines represent
-/// ceiling): ```markdown
+/// ceiling):
+/// ```markdown
 /// ⌈       mCCD  ⌉    ⌈ NRG * mCCD ⌉
```
limemloh, line 1859:
```suggestion
/// ⌈       mCCD  ⌉   ⌈ NRG * mCCD ⌉
```

limemloh (approved):
LGTM

concordium-smart-contract-testing/src/impls.rs
```diff
@@ -84,33 +93,343 @@ impl ChainParameters {
...
+            chain.parameters.euro_per_energy,
+            chain.parameters.micro_ccd_per_euro,
+        )?;
+
+        match (self.block_time, self.block_time_from_external) {
+            (Some(_), true) => return Err(ChainBuilderError::ConflictingBlockTime),
+            (Some(block_time), _) => {
+                chain.parameters.block_time = block_time;
+            }
+            (_, true) => {
```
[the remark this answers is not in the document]
Bargsteen: Good point 😊

concordium-smart-contract-testing/src/impls.rs
Bargsteen: Hmm, they already look aligned on my screen?

PR #98, opened 2023-06-28
Add drop down for smart contract names

### LINKED ISSUE #93: front-end: Make contract name a drop-down based on module
**Description**

Currently the front-end asks the user to input the contract name. We can parse the list of all possible contracts from the module, and offer them in a drop-down. If we have the schema we can also partially recover this information from the schema, if we assume all init functions are in the schema, which technically they don't have to be.

DOBEN:
## Purpose

addresses #93 

## Changes

Add a drop-down element for all available smart contract names.

Users can now select the `smartContractName` that they want to initialize from a drop-down (This is available if the user checked the `Use Module from Step 1` checkbox). The smart contract names are extracted from the WASM module. A valid versioned WASM module needs to be given in step 1 to extract the smart contract names.

### DIFF
--- a/front-end-tools/src/Main.tsx
+++ b/front-end-tools/src/Main.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable no-console */
-import React, { useEffect, useState, ChangeEvent, PropsWithChildren } from 'react';
+import React, { useEffect, useState, ChangeEvent, PropsWithChildren, useCallback, useRef } from 'react';
 import {
     WalletConnectionProps,
     useConnection,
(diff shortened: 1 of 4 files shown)


front-end-tools/src/Main.tsx
```diff
@@ -375,6 +386,32 @@ export default function Main(props: ConnectionProps) {
                                                 const file = hTMLInputElement.files[0];
                                                 const arrayBuffer = await file.arrayBuffer();
 
+                                                // Concordium's tooling create versioned modules e.g. `.wasm.v1` now.
+                                                // Unversioned modules cannot be created by Concordium's tooling anymore.
+                                                // We assume that the uploaded module is a versioned module
+                                                // and remove the versioned 8 bytes at the beginning.
+                                                const wasmModule = await WebAssembly.compile(
+                                                    arrayBuffer.slice(8)
+                                                ).catch((e) => {
```
bisgardo, line 395: Combining `await` and `catch` seems odd. I think you should stick with pure promise `then/catch` or handle error with `try/catch`.

front-end-tools/src/Main.tsx
```diff
@@ -478,29 +515,55 @@ export default function Main(props: ConnectionProps) {
...
+                                                setModuleReferenceError('');
+                                                setModuleReference('');
+                                                setInitName('');
                                                 const checkboxElement = document.getElementById(
                                                     'useModuleReferenceFromStep1'
                                                 ) as HTMLInputElement;
 
+                                                setCheckedBoxElemenChecked(checkboxElement.checked);
+
+                                                const element = document.getElementById(
```
bisgardo, line 527: This doesn't look right. I'm not seeing any elements with that `id` in the project?

But it also isn't the way you're supposed to do it in React. Either use `ref` or store the contents in component state.

front-end-tools/src/Main.tsx
```diff
@@ -89,7 +92,7 @@ export default function Main(props: ConnectionProps) {
         setModuleReference(target.value);
     };
 
-    const changeDropDownHandler = () => {
+    const changeInputParameterDropDownHandler = () => {
```
bisgardo, line 95: Consider using `useCallback` to avoid redundant rerendering of the component(s) that use it.

front-end-tools/src/Main.tsx
```diff
@@ -374,6 +393,33 @@ export default function Main(props: ConnectionProps) {
...
+                                                // We assume that the uploaded module is a versioned module
+                                                // and remove the versioned 8 bytes at the beginning.
+                                                let wasmModule;
+                                                try {
+                                                    wasmModule = await WebAssembly.compile(arrayBuffer.slice(8));
+                                                } catch (e) {
+                                                    setUploadError(
+                                                        `You might have not uploaded a versioned module. Original error: ${
+                                                            (e as Error).message
+                                                        }`
```
abizjak, line 407: We could actually handle this fairly simply.

If the byte array starts with the bytes
```
pub const MAGIC_HASH: [u8; 4] = [0x00, 0x61, 0x73, 0x6D];
```

Then it is a raw Wasm module, without the versioning. Maybe that's worth adding to avoid having this error in case they somehow upload a `.wasm` file.

abizjak (approved):
Looks good, but can we make the drop-down box the same size as the other input boxes?
Screenshot from 2023-06-29 22-19-55

Now it's quite a lot smaller and it looks odd.

PR #101, opened 2023-06-30
Add commands to cargo-concordium to output a JSON template of the schema

### LINKED ISSUE #105: Cargo-concordium: Add support to print a schema template
**Task description**

Add support to output a schema template.

Display traits are now implemented for the schemas.

https://github.com/Concordium/concordium-contracts-common/pull/99

**SubTasks**
- [x] Add `--schema_template_out` flag to the `cargo concordium build` command. It either outputs the JSON template of the schema to a file (if a valid path+file is provided with the flag) or outputs it to the console (if a dash is used aka `-` with the flag)
- [x] Add `cargo concordium schema-template` command (a schema has to be provided either as part of a smart contract module or with the schema flag. Exactly one of the two flags (`--schema` or `--module`) has to be used with this command.) that also outputs the template to a file or console.

DOBEN:
## Purpose

closes #105 

Add support to output a schema template.

## Changes

- Add `--schema_template_out` flag to the `cargo concordium build` command. It either outputs the JSON template of the schema to a file (if a valid `path+file` is provided with the flag) or outputs it to the console (if  a dash is used aka `-` with the flag)
- Add `cargo concordium schema-template` command (a schema has to be provided either as part of a smart contract module or with the schema flag. Exactly one of the two flags (`--schema` or `--module`) has to be used with this command.) that also outputs the template to a file or console.

### DIFF
--- a/cargo-concordium/Cargo.toml
+++ b/cargo-concordium/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cargo-concordium"
-version = "3.1.0"
+version = "3.0.0"
 authors = ["Concordium <developers@concordium.com>"]
 edition = "2021"
 license-file = "../LICENSE"
--- a/cargo-concordium/Cargo.toml
+++ b/cargo-concordium/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cargo-concordium"
-version = "3.0.0"
+version = "3.1.0"
 authors = ["Concordium <developers@concordium.com>"]
 edition = "2021"
 license-file = "../LICENSE"
--- a/cargo-concordium/src/build.rs
+++ b/cargo-concordium/src/build.rs
@@ -27,6 +27,7 @@ use std::{
     cmp::Ordering,
     collections::{BTreeMap, BTreeSet},
     env, fs,
+    io::Write,
     path::{Path, PathBuf},
     process::{Command, Stdio},
 };
@@ -446,8 +447,14 @@ fn write_schema_json(
         fs::create_dir_all(out_dir)
             .context("Unable to create directory for the resulting JSON schemas.")?;
     }
-    std::fs::write(out_path, serde_json::to_string_pretty(&schema_json)?)
-        .context("Unable to write schema output.")?;
+    let mut out_file =
+        std::fs::File::create(out_path).context("Unable to create the output file.")?;
+    write!(
+        &mut out_file,
+        "{}",
+        serde_json::to_string_pretty(&schema_json)?
+    )
+    .context("Unable to write schema json output.")?;
     Ok(())
 }
 
@@ -464,14 +471,17 @@ pub fn write_schema_template(
                 "   Writing the template of the schema to {}.",
                 out.display()
             );
+
             if let Some(out_dir) = out.parent() {
                 fs::create_dir_all(out_dir).context(
                     "Unable to create directory for the resulting template of the schema.",
                 )?;
             }
             // saving the template of the schema to the file
-            std::fs::write(out, format!("{}", schema))
-                .context("Unable to write the template of the schema output.")?;
+            let mut out_file =
+                std::fs::File::create(out).context("Unable to create the output file.")?;
+            write!(&mut out_file, "{}", schema)
+                .context("Unable to write template schema output.")?;
         }
         // printing template of the schema to console
         None => {
@@ -499,7 +509,10 @@ pub fn write_schema_base64(
                     .context("Unable to create directory for the resulting base64 schema.")?;
             }
             // saving the schema base64 representation to the file
-            std::fs::write(out, schema_base64).context("Unable to write schema output.")?;
+            let mut out_file =
+                std::fs::File::create(out).context("Unable to create the output file.")?;
+            write!(&mut out_file, "{}", schema_base64)
+                .context("Unable to write schema base64 output.")?;
         }
         // printing base64 schema to console
         None => {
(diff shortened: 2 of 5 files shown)


cargo-concordium/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "cargo-concordium"
-version = "3.0.0"
+version = "3.1.0"
```
abizjak, line 3: We'll bump the version for release, that's what we do typically.

cargo-concordium/src/build.rs
```diff
@@ -451,6 +451,37 @@ fn write_schema_json(
...
+            );
+            if let Some(out_dir) = out.parent() {
+                fs::create_dir_all(out_dir).context(
+                    "Unable to create directory for the resulting template of the schema.",
+                )?;
+            }
+            // saving the template of the schema to the file
+            std::fs::write(out, format!("{}", schema))
+                .context("Unable to write the template of the schema output.")?;
+        }
```
abizjak, line 475:
```suggestion
            // saving the template of the schema to the file
            let mut out_file = std::fs::File::create(out).context("Unable to create the output file.")?;
            write!(&mut out_file, "{}", schema).context("Unable to write template of the schema output.")?;
        }
```

To avoid creating a string.

cargo-concordium/src/main.rs
```diff
@@ -200,6 +200,56 @@ A schema has to be provided either as part of a smart contract module or with th
...
+    )]
+    SchemaTemplate {
+        #[structopt(
+            name = "out",
+            long = "out",
+            short = "o",
+            default_value = "-",
+            help = "Path and filename to write the converted template representation to or use \
+                    the default value `-` to print the template schema to the console. The path \
+                    has to exist while the file will be created. (expected input: \
```
abizjak, line 219: Why do we not create the path? Don't we actually do that already?

cargo-concordium/src/build.rs
```diff
@@ -451,6 +451,37 @@ fn write_schema_json(
...
+) -> anyhow::Result<()> {
+    match out {
+        // writing the template of the schema to a file
+        Some(out) => {
+            println!(
+                "   Writing the template of the schema to {}.",
+                out.display()
+            );
+            if let Some(out_dir) = out.parent() {
+                fs::create_dir_all(out_dir).context(
```
abizjak, line 468: Does this work correctly for relative paths where `.parent` returns `Some("")`?

cargo-concordium/src/main.rs
```diff
@@ -225,22 +275,32 @@ struct BuildOptions {
...
     )]
-    schema_json_out:   Option<PathBuf>,
+    schema_json_out:     Option<PathBuf>,
+    #[structopt(
+        name = "schema-template-out",
+        long = "schema-template-out",
+        short = "p",
+        help = "Writes the template of the schema to file at specified location or prints the \
+                template of the schema to the console if the value `-` is used. The path has to \
+                exist while the file will be created. (expected input: \
```
abizjak, line 300: I think we are creating the parent directory

abizjak: Can we either change the text, or keep the behaviour consistent.

The behaviour should be consistent across all commands that have `out` flags.

cargo-concordium/src/main.rs
```diff
@@ -552,6 +613,31 @@ pub fn main() -> anyhow::Result<()> {
...
+                .context("Could not get schema.")?;
+
+            if out.as_path() == Path::new("-") {
+                write_schema_template(None, &schema)
+                    .context("Could not print the template of the schema.")?;
+            } else {
+                // A valid path needs to be provided when using the `--out` flag.
+                if out.file_name().is_none() || out.is_dir() {
+                    anyhow::bail!(
+                        "The `--out` flag should point to an existing directory + filename \
```
abizjak, line 632: I'm not sure the existing directory part is true. You create the parent path in `write_schema_tempalte`.

cargo-concordium/src/main.rs
```diff
@@ -631,6 +717,22 @@ fn handle_build(options: BuildOptions, print_schema_info: bool) -> anyhow::Resul
...
                 .context("Could not write JSON schema files.")?;
         }
+        if let Some(schema_template_out) = options.schema_template_out {
+            if schema_template_out.as_path() == Path::new("-") {
+                write_schema_template(None, module_schema)
+                    .context("Could not print the template of the schema.")?;
+            } else {
+                if schema_template_out.file_name().is_none() || schema_template_out.is_dir() {
+                    anyhow::bail!(
+                        "The `--schema-template-out` flag should point to an existing directory + \
```
abizjak, line 727: Same here.

cargo-concordium/src/main.rs
DOBEN: yes, we create the path already. I removed that sentence and also removed the sentence in `SchemaBase64` command.

cargo-concordium/src/build.rs
DOBEN: yes, it works for relative paths:
```
cargo concordium build --schema-template-out ./myTemplate.txt
```

cargo-concordium/src/main.rs
DOBEN: Changed text.

cargo-concordium/src/build.rs
abizjak: And I assume also
```
cargo concordium build --schema-template-out myTemplate.txt
```

abizjak (approved):

cargo-concordium/src/build.rs
DOBEN: yes, that also works.

PR #106, opened 2023-07-04
Get template from moduleSchema

### LINKED ISSUE #94: front-end-tool: Displaying the inputParameters schema to the  user
**Task description**

Currently, the front-end requires the user to input the inputParameter without displaying its format. We can parse the schema of the initParameter from the module (if it has an embedded schema) or from the uploaded schema and display it to the user. This will be possible soon: 
https://github.com/Concordium/concordium-contracts-common/issues/96

- [x] Get embedded schema from module.
- [x] Get the inputParameter template from the schema and display it to the user.

DOBEN:
## Purpose

closes #94 

- Making use of the embedded schemas in the module to simplify the user flow. The front end also displays the schema `template` to the user now.

## Changes

- When the user checks the `Use Module from Step 1` checkbox, the input parameter schema is extracted from the uploaded module (in case the module has an embedded schema). The template of this embedded input parameter schema is shown to the user. No manual upload of the schema is needed anymore in this case.

- The template of the schema is also displayed when the user manually uploads a `schema.bin` file, a corresponding/valid `contract_name` has to be given at the same time for the front end to display the schema.

Scope: Extracting embedded schemas and the template display works from `myContract.wasm.v1`/`schema.bin` files that are created with the current Concordium tooling. It was decided to be out of scope to investigate if it works for very old modules/schemas that were created in the past since it won't add a lot of value (most users want to deploy a module that they currently actively develop)

### DIFF
--- a/front-end-tools/src/Main.tsx
+++ b/front-end-tools/src/Main.tsx
@@ -23,7 +23,7 @@ import { WalletConnectionTypeButton } from './WalletConnectorTypeButton';
 
 import { initialize, deploy } from './writing_to_blockchain';
 
-import { BROWSER_WALLET, REFRESH_INTERVAL } from './constants';
+import { BROWSER_WALLET, REFRESH_INTERVAL, DEFAULT_ARRAY_OBJECT, DEFAULT_JSON_OBJECT } from './constants';
 
 type TestBoxProps = PropsWithChildren<{
     header: string;
@@ -63,7 +63,7 @@ export default function Main(props: ConnectionProps) {
     const [parsingError, setParsingError] = useState('');
     const [smartContractIndexError, setSmartContractIndexError] = useState('');
     const [moduleReferenceError, setModuleReferenceError] = useState('');
-    const [embeddedSchemaError, setEmbeddedSchemaError] = useState('');
+    const [schemaError, setSchemaError] = useState('');
 
     const [accountExistsOnNetwork, setAccountExistsOnNetwork] = useState(true);
     const [moduleReferenceCalculated, setModuleReferenceCalculated] = useState('');
@@ -75,18 +75,18 @@ export default function Main(props: ConnectionProps) {
 
     const [accountBalance, setAccountBalance] = useState('');
     const [inputParameter, setInputParameter] = useState('');
-    const [initName, setInitName] = useState('');
+    const [contractName, setContractName] = useState('');
     const [moduleReference, setModuleReference] = useState('');
     const [cCDAmount, setCCDAmount] = useState('');
     const [base64Module, setBase64Module] = useState('');
-    const [base64Schema, setBase64Schema] = useState('');
+    const [uploadedModuleSchemaBase64, setUploadedModuleSchemaBase64] = useState('');
     const [dropDown, setDropDown] = useState('number');
     const [smartContractIndex, setSmartContractIndex] = useState('');
     const [maxContractExecutionEnergy, setMaxContractExecutionEnergy] = useState('');
-    const [checkedBoxElemenChecked, setCheckedBoxElemenChecked] = useState(false);
+    const [useModuleFromStep1, setUseModuleFromStep1] = useState(false);
     const [contracts, setContracts] = useState<string[]>([]);
+    const [displayContracts, setDisplayContracts] = useState<string[]>([]);
     const [inputParameterTemplate, setInputParameterTemplate] = useState('');
-    const [inputParamterTypeSchema, setInputParamterTypeSchema] = useState('');
     const [embeddedModuleSchemaBase64, setEmbeddedModuleSchemaBase64] = useState('');
 
     const [isWaitingForTransaction, setWaitingForUser] = useState(false);
@@ -100,17 +100,7 @@ export default function Main(props: ConnectionProps) {
     const inputParameterTextAreaRef = useRef(null);
     const useModuleReferenceFromStep1Ref = useRef(null);
     const moduleReferenceRef = useRef(null);
-
-    const defaultJSONObject = {
-        myStringField: 'FieldValue',
-        myNumberField: 4,
-        myArray: [1, 2, 3],
-        myObject: {
-            myField1: 'FieldValue',
-        },
-    };
-
-    const defaultArrayObject = 'Examples: \n\n[1,2,3] or \n\n["abc","def"] or \n\n[{"myFieldKey":"myFieldValue"}]';
+    const inputParameterFieldRef = useRef(null);
 
     function arraysEqual(a: Uint8Array, b: Uint8Array) {
         if (a === b) return true;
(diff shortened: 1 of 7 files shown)


shjortConcordium:
The new SDK have been released now :upside_down_face:
(So we should be able to avoid copying around the SDK)

DOBEN:
> The new SDK have been released now upside_down_face (So we should be able to avoid copying around the SDK)

Updated the SDK :rocket:

shjortConcordium (approved):
LGTM

front-end-tools/src/Main.tsx
```diff
@@ -690,67 +783,92 @@ export default function Main(props: ConnectionProps) {
...
+                                        {inputParameterTemplate && (
                                             <div className="actionResultBox">
-                                                Schema in base64:
-                                                <div>{base64Schema.toString().slice(0, 30)} ...</div>
+                                                Input Parameter Template:
+                                                <pre>
+                                                    {JSON.stringify(
+                                                        JSON.parse(JSON.parse(inputParameterTemplate)),
+                                                        undefined,
+                                                        2
```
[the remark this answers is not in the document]
DOBEN: Simplified the `JSON.parsing`. Now there needs to be no double `JSON.parsing` anymore.

front-end-tools/src/Main.tsx
```diff
@@ -588,13 +660,29 @@ export default function Main(props: ConnectionProps) {
...
                                         />
                                         <span>{' Use Module from Step 1'}</span>
                                     </label>
                                 </div>
+                                {useModuleFromStep1 && (
+                                    <>
+                                        <br />
+                                        <div>
+                                            This checkbox autofilled the `moduleReference`, the `contractNames`, and the
+                                            `embeddedInputParameterSchema` from the above module.
```
abizjak, line 677: I don't think this backtick works very well. It looks ugly. Perhaps use https://getbootstrap.com/docs/4.0/content/code/#inline-code instead?
Screenshot from 2023-07-11 20-15-13

front-end-tools/src/Main.tsx
```diff
@@ -588,13 +660,29 @@ export default function Main(props: ConnectionProps) {
...
+                                        </div>
+                                        <div>
+                                            If you want to load a new module from step 1, `uncheck` and `check` this box
```
abizjak, line 680: Also here I think you want to just use bold or italics to highlight the action.

front-end-tools/src/Main.tsx
```diff
@@ -588,13 +660,29 @@ export default function Main(props: ConnectionProps) {
...
                                             }}
                                         />
                                         <span>{' Use Module from Step 1'}</span>
                                     </label>
                                 </div>
+                                {useModuleFromStep1 && (
+                                    <>
+                                        <br />
+                                        <div>
+                                            This checkbox autofilled the `moduleReference`, the `contractNames`, and the
```
abizjak, line 676: It's unclear from the user interface what "contractNames" refers to. The UI element has the header "Smart contract name". Similarly "moduleReference" is "Module reference".

front-end-tools/src/Main.tsx
```diff
@@ -276,6 +294,43 @@ export default function Main(props: ConnectionProps) {
...
+                    2
+                );
+
+                template = displayTypeSchemaTemplate(inputParamterTypeSchemaBuffer);
+
+                setInputParameterTemplate(template);
+            } catch (e) {
+                if (useModuleFromStep1) {
+                    setSchemaError(
+                        `Could not get embedded input parameter schema from the uploaded module. Uncheck "Use Module from Step 1" checkbox to upload manually a schema. Original error: ${e}`
```
abizjak, line 317:
```suggestion
                        `Could not get embedded input parameter schema from the uploaded module.\nUncheck "Use Module from Step 1" checkbox to manually upload a schema. Original error: ${e}`
```

front-end-tools/src/constants.ts
```diff
@@ -4,3 +4,14 @@ import moment from 'moment';
 export const REFRESH_INTERVAL = moment.duration(5, 'seconds');
 
 export const BROWSER_WALLET = ephemeralConnectorType(BrowserWalletConnector.create);
+
+export const DEFAULT_JSON_OBJECT = {
```
abizjak, line 8: Can you elaborate on what this is meant for?

front-end-tools/src/constants.ts
```diff
@@ -4,3 +4,14 @@ import moment from 'moment';
...
+    myStringField: 'FieldValue',
+    myNumberField: 4,
+    myArray: [1, 2, 3],
+    myObject: {
+        myField1: 'FieldValue',
+    },
+};
+
+export const DEFAULT_ARRAY_OBJECT = 'Examples: \n\n[1,2,3] or \n\n["abc","def"] or \n\n[{"myFieldKey":"myFieldValue"}]';
```
abizjak, line 17: This is an oddly named constant, if I compare it to DEFAULT_JSON_OBJECT. This is a string. Could you document the two please.

abizjak (approved):
Functionality wise looks good. But please address the UI comments I have.

It would be better if we could automatically refresh the schema template if I upload a new module, but that can be done in a followup work.

front-end-tools/src/constants.ts
DOBEN: This is the example JSON object that is shown in the textfield when the user has no `embedded schema` in the module or does not want to use the embedded schema (e.g. if the checkbox "Use module from step 1" is unchecked).

front-end-tools/src/constants.ts
DOBEN: Added the following comment:

// This is the example array that is shown in the input parameter textarea as a placeholder when the user has no embedded schema in the module 
// or does not want to use the embedded schema (meaning if the checkbox "Use module from step 1" is unchecked).

DOBEN: Renamed it to:
`EXAMPLE_ARRAYS`

front-end-tools/src/constants.ts
DOBEN: Renamed it to:
`EXAMPLE_JSON_OBJECT`

DOBEN:
> It would be better if we could automatically refresh the schema template if I upload a new module, but that can be done in a followup work.

I am not fully convinced that is the best behavior when we upload a new module that the `schema template`, `smart contract names`, `module reference` are automatically refreshed or get changed. I want to either refresh all three things or none of them automatically to avoid confusion.

- I try to minimize any change to the input parameter values that happen automatically since it can be confusing for a new user (e.g. a new user might not know/understand that the embedded schema comes from the module that was uploaded in step 1). When input values are changed automatically then also there is a risk that you `initialize` a transaction with some values that you have not explicitly set (because they have changed to what you have set).
When the user has to explicitly check the `use module from step 1`, I hope the user is more aware that some input values are now automatically changed/updated.

abizjak:
> > It would be better if we could automatically refresh the schema template if I upload a new module, but that can be done in a followup work.
> 
> I am not fully convinced that is the best behavior when we upload a new module that the `schema template`, `smart contract names`, `module reference` are automatically refreshed or get changed. I want to either refresh all three things or none of them automatically to avoid confusion.
> 
> I try to minimize any change to the input field (or text area) values that happen automatically since it can be confusing for a new user (e.g. a new user might not know/understand that the embedded schema comes from the module that was uploaded in step 1). When input values are changed automatically then also there is a risk that you `initialize` a transaction with some values that you have not explicitly set (because they have changed to what you have set). When the user has to explicitly check the `use module from step 1`, I hope the user is more aware that some values are now automatically changed/updated.
> 
> Let's discuss it and maybe we can work on an `expert` mode for the front-end that automatically refreshes values.

That sounds reasonable a design. Instead of automatically updating then, just making sure that it is visible that potential dependencies changed would help.

But as I wrote before, this can be done in followup work.

PR #110, opened 2023-07-14
Re-export types needed for the integration testing library.

DOBEN:
## Purpose

Re-export types needed for the integration testing library.

## Changes

Re-export types needed for account public key query and account signature verification.

### DIFF
--- a/concordium-smart-contract-testing/src/constants.rs
+++ b/concordium-smart-contract-testing/src/constants.rs
@@ -28,9 +28,9 @@ pub(crate) fn contract_instance_query_account_keys_return_cost(num_keys: u32) ->
 
 /// Cost **in energy** of verification of an ed25519 signature.
 /// This should match the cost of
-/// [`verify_ed22519_cost`](concordium_smart_contract_engine::constants::
-/// verify_ed25519_cost) except the latter is the cost in interpreter energy,
-/// and this on is in [`Energy`].
+/// [`verify_ed22519_cost`](concordium_smart_contract_engine::constants::verify_ed25519_cost)
+/// except the latter is the cost in interpreter energy, and this on is in
+/// [`Energy`].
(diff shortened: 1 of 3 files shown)


concordium-smart-contract-testing/src/constants.rs
```diff
@@ -28,9 +28,9 @@ pub(crate) fn contract_instance_query_account_keys_return_cost(num_keys: u32) ->
 
 /// Cost **in energy** of verification of an ed25519 signature.
 /// This should match the cost of
-/// [`verify_ed22519_cost`](concordium_smart_contract_engine::constants::verify_ed25519_cost)
-/// except the latter is the cost in interpreter energy, and this on is in
-/// [`Energy`].
+/// [`verify_ed22519_cost`](concordium_smart_contract_engine::constants::
+/// verify_ed25519_cost) except the latter is the cost in interpreter energy,
```
limemloh, line 32: Seems weird if the formatter changed this. Did you add a newline here by mistake?

concordium-smart-contract-testing/src/types.rs
```diff
@@ -36,7 +36,7 @@ pub(crate) struct ChainParameters {
     /// Euro per Energy ratio.
     // This is not public because we ensure a reasonable value during the construction of the
     // [`Chain`].
-    pub(crate) euro_per_energy: ExchangeRate,
+    pub(crate) euro_per_energy:    ExchangeRate,
```
limemloh, line 39: Also here, did we change the formatting?

limemloh (approved):

concordium-smart-contract-testing/src/constants.rs
abizjak: I think this will break the link, so this should be reverted or fixed some other way.

abizjak (approved):
If you can fix that link that was split into two lines then this looks good.

PR #113, opened 2023-07-19
Fix the order of expected arguments to match the node.

abizjak:
## Purpose

Fix the order of arguments to check signature so that it matches the node.

## Checklist

      hard-to-understand areas.


DOBEN (approved):
Thanks, it works brilliantly now.

PR #114, opened 2023-07-26
Fix check if module is deployed already in step 2

DOBEN:
## Purpose

- Fix the `isModuleReferenceAlreadyDeployedStep2` check. The current hosted front end does not allow a user to immediately jump to step 2 and initialize a module by inputting a module reference e.g. obtained from looking it up on a block explorer.

## Changes

- `IsModuleReferenceAlreadyDeployedStep2` uses the module reference from the input field in step 2. 
- `isModuleReferenceAlreadyDeployedStep1` uses the module reference from the calculation in step 1.

### DIFF
--- a/front-end-tools/CHANGELOG.md
+++ b/front-end-tools/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 2.1.0
 
-- Fix the `isModuleReferenceAlreadyDeployedStep2` check. `isModuleReferenceAlreadyDeployedStep2` uses the module reference from the input field in step 2 while the `isModuleReferenceAlreadyDeployedStep1` uses the module reference from the calculation in step 1.
+- Fix check if module is already deployed in step 2.
 
 ## 2.0.0
 
--- a/front-end-tools/CHANGELOG.md
+++ b/front-end-tools/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased changes
 
+## 2.1.0
+
+- Fix the `isModuleReferenceAlreadyDeployedStep2` check. `isModuleReferenceAlreadyDeployedStep2` uses the module reference from the input field in step 2 while the `isModuleReferenceAlreadyDeployedStep1` uses the module reference from the calculation in step 1.
+
 ## 2.0.0
 
 - Add support for getting the embedded schema from the module.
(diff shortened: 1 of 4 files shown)


front-end-tools/src/Main.tsx
```diff
@@ -283,17 +284,34 @@ export default function Main(props: ConnectionProps) {
...
                     }
                 })
                 .catch(() => {
-                    setModuleReferenceAlreadyDeployed(false);
+                    setIsModuleReferenceAlreadyDeployedStep1(false);
                 });
         }
     }, [connection, account, client, moduleReferenceCalculated]);
 
+    useEffect(() => {
```
schwartz-concordium, line 298: Isn’t it possible only to have one `moduleReferenceAlreadyDeployed` state, which could have a effect changing listening to both `moduleReferenceCalculated`and `moduleReference `? 

The warning should be shown either way regardless of where the user provides the module (from file or string).

front-end-tools/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Unreleased changes
 
+## 2.1.0
+
+- Fix the `isModuleReferenceAlreadyDeployedStep2` check. `IsModuleReferenceAlreadyDeployedStep2` uses the module reference from the input field in step 2 while the `isModuleReferenceAlreadyDeployedStep1` uses the module reference from the calculation in step 1.
```
schwartz-concordium, line 5:
```suggestion
- Fix the `isModuleReferenceAlreadyDeployedStep2` check. `isModuleReferenceAlreadyDeployedStep2` uses the module reference from the input field in step 2 while the `isModuleReferenceAlreadyDeployedStep1` uses the module reference from the calculation in step 1.
```

front-end-tools/src/Main.tsx
DOBEN: So how should we set `moduleReferenceAlreadyDeployed` if `moduleReferenceCalculatedand is NOT deployed` but `moduleReference is deployed` or vice versa?

schwartz-concordium (approved):

front-end-tools/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Unreleased changes
 
+## 2.1.0
+
+- Fix the `isModuleReferenceAlreadyDeployedStep2` check. `isModuleReferenceAlreadyDeployedStep2` uses the module reference from the input field in step 2 while the `isModuleReferenceAlreadyDeployedStep1` uses the module reference from the calculation in step 1.
```
abizjak, line 5: The changelog for a tool should typically reflect user understandable or observable changes. 

So I'd avoid mentioning specific function names which have no meaning outside of the codebase.

abizjak (approved):

PR #115, opened 2023-07-31
Add warning boxes for some edge cases

DOBEN:
## Purpose

Collection of improvements to give better feedback to the user in some edge cases that likely are caused by the user's wrong input.

## Changes

- Add a warning box if module references in step 1 and step 2 are not the same. 
- Add a warning box if embedded schema indicates an input parameter but no input parameter is provided.
- Add check if inputted module reference has the correct length.
- Module reference input field is now disabled if the user checked `Use Module From Step 1` check box to avoid some edge cases.
- Revised some text displayed to the user

### DIFF
--- a/front-end-tools/src/Main.tsx
+++ b/front-end-tools/src/Main.tsx
@@ -73,7 +73,7 @@ export default function Main(props: ConnectionProps) {
     const [moduleReferenceDeployed, setModuleReferenceDeployed] = useState('');
 
     const [shouldWarnDifferenceModuleReferences, setShouldWarnDifferenceModuleReferences] = useState(false);
-    const [shouldWarnInputPramaterInSchemaIgnored, setShouldWarnInputPramaterInSchemaIgnored] = useState(false);
+    const [shouldWarnInputParameterInSchemaIgnored, setShouldWarnInputParameterInSchemaIgnored] = useState(false);
 
     const [txHashDeploy, setTxHashDeploy] = useState('');
     const [txHashInit, setTxHashInit] = useState('');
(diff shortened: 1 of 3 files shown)


front-end-tools/src/Main.tsx
```diff
@@ -71,6 +72,9 @@ export default function Main(props: ConnectionProps) {
     const [isModuleReferenceAlreadyDeployedStep2, setIsModuleReferenceAlreadyDeployedStep2] = useState(false);
     const [moduleReferenceDeployed, setModuleReferenceDeployed] = useState('');
 
+    const [shouldWarnDifferenceModuleReferences, setShouldWarnDifferenceModuleReferences] = useState(false);
+    const [shouldWarnInputPramaterInSchemaIgnored, setShouldWarnInputPramaterInSchemaIgnored] = useState(false);
```
schwartz-concordium, line 76:
```suggestion
    const [shouldWarnInputParameterInSchemaIgnored, setShouldWarnInputParameterInSchemaIgnored] = useState(false);
```

schwartz-concordium (changes requested):

front-end-tools/src/Main.tsx
```diff
@@ -978,14 +1032,30 @@ export default function Main(props: ConnectionProps) {
...
+                                )}
                                 {!txHashInit && transactionErrorInit && (
                                     <div className="alert alert-danger" role="alert">
                                         Error: {transactionErrorInit}.
                                     </div>
                                 )}
                                 {txHashInit && (
                                     <>
-                                        Transaction hash (May take a moment to finalize): {}
+                                        <div>Transaction hash:</div>
```
abizjak, line 1053: Same here I think.

schwartz-concordium (approved):

PR #118, opened 2023-08-03
Add support for the `--manifest-path` flag

DOBEN:
## Purpose

Add support for the `--manifest-path` flag.

## Changes
Add support for the `cargo` native flag `--manifest-path`. The following commands are now working:

```cargo concordium build -- --manifest-path ./path/to/my/project/Cargo.toml```
```cargo concordium test -- --manifest-path ./path/to/my/project/Cargo.toml```

### DIFF
--- a/cargo-concordium/src/build.rs
+++ b/cargo-concordium/src/build.rs
@@ -58,7 +58,10 @@ fn get_crate_metadata(cargo_args: &[String]) -> anyhow::Result<Metadata> {
         }
         Some(p) => {
             // If a `--manifest-path` is provided, look for the `Cargo.toml` file there.
-            cmd.manifest_path(p.trim_start_matches("--manifest-path="));
+            cmd.manifest_path(
+                p.strip_prefix("--manifest-path=")
+                    .context("Incorrect `--manifest-path` flag.")?,
+            );
         }
         None => {
             // If NO `--manifest-path` is provided, look for the `Cargo.toml`
@@ -746,13 +749,13 @@ pub(crate) fn build_and_run_integration_tests(
     build_options: BuildOptions,
     test_targets: Vec<String>,
 ) -> anyhow::Result<()> {
-    let cargo_args = build_options.cargo_args.clone();
+    let cargo_args = &build_options.cargo_args.clone();
 
     // Build the module in the same way as `cargo concordium build`, except that
     // schema information shouldn't be printed.
     crate::handle_build(build_options, false)?;
 
-    let metadata = get_crate_metadata(&cargo_args)?;
+    let metadata = get_crate_metadata(cargo_args)?;
 
     let mut cargo_test_args = vec!["test"];
 
--- a/cargo-concordium/src/build.rs
+++ b/cargo-concordium/src/build.rs
@@ -116,7 +116,7 @@ pub fn build_contract(
         }
     };
 
-    let (target_dir, metadata) = if cargo_args.contains(&String::from("--manifest-path")) {
+    let metadata = if cargo_args.contains(&String::from("--manifest-path")) {
         // If a `--manifest-path` flag is provided use this path as the package root to
         // build the project
         let index = cargo_args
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cargo-concordium/src/build.rs [hunk 1]
```diff
@@ -798,16 +854,34 @@ pub(crate) fn build_and_run_integration_tests(
 /// The `seed` argument allows for providing the seed to instantiate a random
 /// number generator. If `None` is given, a random seed will be sampled.
 pub fn build_and_run_wasm_test(extra_args: &[String], seed: Option<u64>) -> anyhow::Result<bool> {
-    let metadata = MetadataCommand::new()
-        .exec()
-        .context("Could not access cargo metadata.")?;
+    let mut args = std::env::args().skip_while(|val| !val.starts_with("--manifest-path"));
```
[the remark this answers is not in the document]
abizjak: Can we please do it like this, since that means that the function is self-contained, and does not read environment variables.

cargo-concordium/src/build.rs
```diff
@@ -332,16 +350,34 @@ pub fn build_contract_schema<A>(
...
+        }
+        Some(p) => {
+            // If a `--manifest-path` is provided use this path as the package root to
+            // build the schema.
+            cmd.manifest_path(p.trim_start_matches("--manifest-path="));
+        }
+        None => {}
+    };
+
+    let metadata = cmd.exec().context("Could not access cargo metadata.")?;
```
abizjak, line 373: Could we please factor this into a function
```rust
fn get_crate_metadata(extra_args: &[String]) -> anyhow::Result<Metadata> {
    let mut args = extra_args.iter().skip_while(|val| !val.starts_with("--manifest-path"));
    let mut cmd = MetadataCommand::new();
    match args.next() {
        Some(ref p) if p == "--manifest-path" => {
            // If a `--manifest-path` is provided use this path as the package root to
            // build the schema.
            cmd.manifest_path(args.next().context(
                "The argument '--manifest-path <manifest-path>' requires a value but none was \
                 supplied.",
            )?);
        }
        Some(p) => {
            // If a `--manifest-path` is provided use this path as the package root to
            // build the schema.
            cmd.manifest_path(p.trim_start_matches("--manifest-path="));
        }
        None => {}
    };

    cmd.exec().context("Could not access cargo metadata.")
}
```
Since it is repeated multiple places.

cargo-concordium/src/build.rs
```diff
@@ -798,16 +854,34 @@ pub(crate) fn build_and_run_integration_tests(
...
+            // test the project.
+            cmd.manifest_path(args.next().context(
+                "The argument '--manifest-path <manifest-path>' requires a value but none was \
+                 supplied.",
+            )?);
+        }
+        Some(p) => {
+            // If a `--manifest-path` is provided use this path as the package root to
+            // test the project.
+            cmd.manifest_path(p.trim_start_matches("--manifest-path="));
```
abizjak, line 872: Actually this is a bit wrong, in the sense that it will do the wrong thing if they happen to write `--manifest-pathblabh` as an argument.

It would be better to do
```rust
p.strip_prefix("--manifest-path=").context("Incorrect manifest-path flag.")?
```

DOBEN: cargo run concordium test -- --manifest-pathdsaf

```
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/cargo-concordium concordium test -- --manifest-pathdsaf`
Error: Could not build and run tests.

Caused by:
    0: Could not access cargo metadata.
    1: `cargo metadata` exited with an error: error: unexpected argument '--manifest-pathdsaf' found
       
         note: argument '--manifest-path' exists
       
       Usage: cargo metadata --format-version <VERSION> --manifest-path <PATH>
       
       For more information, try '--help'.
```

It seems to give you some output that the flag name is wrong and should be corrected

cargo-concordium/src/build.rs
```diff
@@ -721,14 +746,13 @@ pub(crate) fn build_and_run_integration_tests(
     build_options: BuildOptions,
     test_targets: Vec<String>,
 ) -> anyhow::Result<()> {
+    let cargo_args = build_options.cargo_args.clone();
```
abizjak, line 749:
```suggestion
    let cargo_args = &build_options.cargo_args;
```

abizjak (approved):

cargo-concordium/src/build.rs
DOBEN: The function one line `crate::handle_build()` below moves the value, so I had to clone it here.

PR #123, opened 2023-10-12
Reading/Writing Features

### LINKED ISSUE #127: Add read from (and write to) smart contract feature to hosted deploy/initialize front end
**Task description**

Our hosted deploy/initialize front-end would benefit from having capabilities to read/write to arbitrary smart contracts to e.g. check if a smart contract protocol has been deployed/initialized correctly.

- Add read from smart contract feature
- Add write to smart contract feature

DOBEN:
## Purpose

closes #127 

- Add a feature to read from smart contract
- Add a feature to write to smart contract

## Changes

- Add a box to read from the smart contract (either the user can manually fill in all input fields of that box or the user can only input the smart contract index and derive most of the values by checking the "Derive From Smart Contract Index" checkbox)
- Add a box to write to the smart contract  (either the user can manually fill in all input fields of that box or the user can only input the smart contract index and derive most of the values by checking the "Derive From Smart Contract Index" checkbox)


front-end-tools/src/Main.tsx
soerenbf, line 1: I would consider doing the following to make this easier to maintain:

- Separate steps into individual components
  - Keep as much state as possible local to the individual component
  - Would also split code into separate files/folders (by component), as it gets hard to clearly distinguish what fits together logically in large files like these.
- Add a forms library to help structure validation of user input and associated errors (I can recommend [react-hook-form](https://react-hook-form.com/)
  - This should also allow you to rely less on `useEffect`, which - especially when combined with state setters - quickly become hard to grasp the effects of.
- If you have any global state that exists throughout the lifetime of the application, it would make sense to introduce a global store (can just be a react context defined at application root).
  - It seems like the only data global to the application is the wallet connection though, and it might be easier to pass the connection props down. You decide 😄

front-end-tools/src/reading_from_blockchain.ts
soerenbf, line 1: Instead of allowing `T | undefined` unions in parameters, I would validate user input/connection state prior to invocation of these functions.

front-end-tools/src/writing_to_blockchain.ts
soerenbf, line 1: Same here.

soerenbf (commented):
Here are some things to consider.

- No wallet connect support? Is this intentional or something we plan to introduce (since we use `@concordium/react-components`)?
- As users can now do more stuff than deploy/init contracts, it would make sense to think about user flows (in my opinion)
  - I would also introduce a router (react-router), as it lets you be more declarative about individual sections of the application, with the added benefit of persisting the location in the url along with associated state. This will allow users to use browser navigation (back/forward) to navigate the app + link to selected parts of the application, which might be useful.

#### Possible application routes (marked with bold)
- **/ (root = Deploy/declare module)**
  - **/module/:module-reference**
    - Directly, in which case we prompt for module ref
    - As a result of module deployment, assume users would like to init instance of module
- **/instance**
  - Declare contract details (can we just always derive contract name and possible entrypoints if possible?)
  - **/instance/:index/read (do we need subindex?)**
  - **/instance/:index/write (do we need subindex?)**

I realize these changes are quite substantial, but will improve usability a lot.

front-end-tools/src/Main.tsx
```diff
@@ -1081,19 +1445,974 @@ export default function Main(props: ConnectionProps) {
...
+                                                            return data + String.fromCharCode(byte);
+                                                        }, '')
+                                                    );
+                                                    setUploadedModuleSchemaBase64Read(schema);
+                                                    console.log(schema);
+                                                } else {
+                                                    setUploadErrorRead('Upload schema file is undefined');
+                                                }
+                                            }}
+                                        />
```
soerenbf, line 1656: Here also it would be more suitable with another solution in my opinion:

```tsx
function validateResult(input: string | Error): true | string { // returns validation message if invalid
    ...
}
...
<Controller
    name="file"
    rules={{ 
        required:"Schema file is required",
        validate: validateResult // Validates whether an error/success result was produced (maybe not needed)?
    }}
    render={(field) =>
        <Form.Control
            type="file"
            accept=".bin"
            {...field}
            onChange={async (e) => {
                const files = e.target.value; // or whatever works
                                                
                if (files === null || files?.length === 0) {
                    field.onChange(undefined); // will trigger "required" validation rule
                    return;
                }
                                                
                try {
                     const file = files[0];
                     const arrayBuffer = await file.arrayBuffer();

                     const schema = btoa(
                         new Uint8Array(arrayBuffer).reduce((data, byte) => {
                             return data + String.fromCharCode(byte);
                         }, '')
                     );
                                                
                     field.onChange(schema);
                 } catch (e) {
                     field.onChange(e); // If it'll never fail, the try/catch and the associated "validate" rule can be removed
                 }
             }}
         />
     }
/>
```

front-end-tools/src/Main.tsx
```diff
@@ -1081,19 +1445,974 @@ export default function Main(props: ConnectionProps) {
...
+                                            hasInputParameterRegister.onChange(e);
+
+                                            setParsingErrorRead(undefined);
+                                            attributeForm.setValue('inputParameterType', undefined);
+                                            attributeForm.setValue('inputParameter', undefined);
+                                            setEntryPointTemplateReadFunction(undefined);
+                                            setSchemaError({
+                                                ...schemaError,
+                                                readFunction: undefined,
+                                            });
```
soerenbf, line 1704: I would move the parts needed into an effect (`useEffect`) instead.

front-end-tools/src/Main.tsx
```diff
@@ -1081,19 +1445,974 @@ export default function Main(props: ConnectionProps) {
...
+                                                    onChange={(e) => {
+                                                        const inputParameterRegister = attributeForm.register(
+                                                            'inputParameter',
+                                                            { required: true }
+                                                        );
+
+                                                        inputParameterRegister.onChange(e);
+
+                                                        setParsingErrorRead(undefined);
+                                                    }}
```
soerenbf, line 1782: Remove this, and if necessary move `setParsingErrorRead` into an effect.

front-end-tools/src/Main.tsx
```diff
@@ -1081,19 +1445,974 @@ export default function Main(props: ConnectionProps) {
...
+                                                                JSON.parse(target.value);
+                                                            } catch (e) {
+                                                                setParsingErrorRead((e as Error).message);
+                                                                return;
+                                                            }
+                                                            attributeForm.setValue('inputParameter', target.value);
+                                                        }}
+                                                    >
+                                                        {getArrayExample(entryPointTemplateReadFunction)}
+                                                    </textarea>
```
soerenbf, line 1820: Wrap in `Controller` instead to do custom `onChange` logic.

front-end-tools/src/Main.tsx
```diff
@@ -1081,19 +1445,974 @@ export default function Main(props: ConnectionProps) {
...
+                                                            const target = event.target as HTMLTextAreaElement;
+
+                                                            try {
+                                                                JSON.parse(target.value);
+                                                            } catch (e) {
+                                                                setParsingErrorRead((e as Error).message);
+                                                                return;
+                                                            }
+                                                            attributeForm.setValue('inputParameter', target.value);
+                                                        }}
```
soerenbf, line 1836: Can we not just consolidate with the array version of this? What's the difference?

front-end-tools/src/Main.tsx
```diff
@@ -1081,19 +1445,974 @@ export default function Main(props: ConnectionProps) {
...
+
+                                        promise
+                                            .then((value) => {
+                                                setReturnValue(value);
+                                            })
+                                            .catch((err: Error) => setReadError((err as Error).message));
+                                    })}
+                                >
+                                    Read Smart Contract
+                                </Button>
```
soerenbf, line 1888: Idiomatic way to do this would be to:

```tsx
const {handleSubmit} = useForm(...);
function onSubmit(values: FormValues) { // or whatever the type of the form is?
...
}
...
<form onSubmit={handleSubmit(onSubmit)}>
    ... // form fields
    <Button type="submit" ...>...</Button>
</form>
```

soerenbf (commented):
Here are some pointers with regards to using react-hook-form in the intended way:

I think if you try to separate your flows into separate components it'll be easier to get an overview of what state belongs to the individual parts of the application. I would probably prioritise this before changing the form stuff, but you do what works for you! 😄

front-end-tools/src/Main.tsx
```diff
@@ -1081,19 +1445,974 @@ export default function Main(props: ConnectionProps) {
...
+                                    contractInstanceInfo.contractName !== undefined ? (
+                                        <Form.Group className="col-md-4 mb-3">
+                                            <Form.Label>Smart Contract Name</Form.Label>
+                                            <Form.Control
+                                                value={
+                                                    contractInstanceInfo?.contractName
+                                                        ? contractInstanceInfo.contractName
+                                                        : 'undefined'
+                                                }
+                                                disabled
```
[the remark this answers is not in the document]
DOBEN: In terms of styling, I think the `disabled` is more clear since it greys out the field and doesn't allow the user to click into the field (the boarders are not highlighted or focused when clicking into the field). The `read-only`  styling allows the user to click into the field and focus on the field and since the field is not greyed out it looks like as if it could be edited by the user.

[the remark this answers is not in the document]
soerenbf: Hmm.. in bootstrap 4 `readonly` would render style like the disabled of bootstrap 5.

Anyway, as I recall disabled inputs are excluded from the form values (native browser form behaviour, not react-hook-form exclusive), so in my opinion the correct way would still be to mark them as `readonly`.

In the [bootstrap docs for disabled form controls](https://getbootstrap.com/docs/5.3/forms/form-control/#disabled), there is an example marking an input as `disabled readonly`, which makes the text selectable (which is also not possible with just `disabled`. Maybe this will include values in form submissions as well?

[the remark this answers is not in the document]
DOBEN: The idea is to `disable` the input field, so users cannot fill in the the value and submit it via the form. The smart contract name was derived from the smart contract index. It should be most obviously that this field cannot be filled in. `Disabled` styling is pretty clear in that sense by making the field greyed out that is why I would lick to stick with the `disabled` styling.

DOBEN:
> * No wallet connect support? Is this intentional or something we plan to introduce (since we use `@concordium/react-components`)?

The mobile wallets do not support `deploy/init` transactions at the moment. That is why walletConnect is not support at the moment. It will be added as soon we have these features in the mobile wallets.

DOBEN:
> As users can now do more stuff than deploy/init contracts, it would make sense to think about user flows (in my opinion)
> 
>     * I would also introduce a router (react-router), as it lets you be more declarative about individual sections of the application, with the added benefit of persisting the location in the url along with associated state. This will allow users to use browser navigation (back/forward) to navigate the app + link to selected parts of the application, which might be useful.

Based on some discussion, adding a router would make more sense in a separate PR. In my opinion, the read/write box should be visible at the page at all time since it can be useful before as well as after the deployment/init steps to do some reading/writing to the blockchain to potentially even unrelated contract addresses.

The step 1 and step 2 could be implemented via navigating them from step 1 to step 2 on separate pages. I keep this discussion for future PRs/issues so.

front-end-tools/src/Main.tsx
DOBEN: Arrays and objects are both treated as JSON objects. 

The difference between both is that we show different examples to the user e.g.
```
export const EXAMPLE_JSON_OBJECT = {
    myStringField: 'FieldValue',
    myNumberField: 4,
    myArray: [1, 2, 3],
    myObject: {
        myField1: 'FieldValue',
    },
};
```
vs.
```
export const EXAMPLE_ARRAYS = 'Examples: \n\n[1,2,3] or \n\n["abc","def"] or \n\n[{"myFieldKey":"myFieldValue"}]';
```

People can then start to edit either the array example or the object example that they choose.

front-end-tools/src/Main.tsx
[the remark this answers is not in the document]
soerenbf: I agree the styling for disabled is as it should be, however I was just trying to make the point that if you rely on the value from the disabled input field on the form submission values, it will not be available:

```ts
const {handleSubmit, register} = useForm();
function onSubmit({field1, field2}) {
    // field1 is defined
    // field2 is undefined
}
...
<form onSubmit={handleSubmit(onSubmit)}>
    <input {...register('field1')} />
    <input {...register('field2')} diasbled />
</form>
```

Does that make sense?

Anyway if you don't use the value from the input field it won't matter.

soerenbf:
> Based on some discussion, adding a router would make more sense in a separate PR. In my opinion, the read/write box should be visible at the page at all time since it can be useful before as well as after the deployment/init steps to do some reading/writing to the blockchain to potentially even unrelated contract addresses.
> 
> The step 1 and step 2 could be implemented via navigating from step 1 to step 2 on separate pages. I keep this discussion for future PRs/issues so. Some state would need to be shared between step 1 and step 2.

Makes sense. With regards to the 2-step flow, it's possible to use browser location state, which can also be restored on back/forward browser navigation.

Anyway it would be nice if you could add an issue for this refactoring 🙂

front-end-tools/src/Main.tsx
soerenbf: Actually.. I don't know the best way to do this. The problem is, that you need the value of the original field (the inner field), to stay the same, while deriving something from the value whenever it changes. I haven't tested this, but what I think I would do would be:

```tsx
type DerivedInput<O, D> = {
    original: O;
    derived: D;
};

function validateFileInput(input: DerivedInput<FileInput, Base64String | Error>): true | string {
    ...
}
...
<Controller
    name="file"
    rules={{ 
        validate: validateFileInput // Validates whether an error/success result was produced (maybe not needed)?
    }}
    render={(field) =>
        <Form.Control
            type="file"
            accept=".bin"
            {...field}
            value={field.value.original}
            onChange={async (e) => {
                const files = e.target.value; // or whatever works
                                                
                if (files === null || files?.length === 0) {
                    field.onChange({original: files, derived: new Error('No file found in input')});
                    return;
                }
                                                
                try {
                     const file = files[0];
                     const arrayBuffer = await file.arrayBuffer();

                     const schema = btoa(
                         new Uint8Array(arrayBuffer).reduce((data, byte) => {
                             return data + String.fromCharCode(byte);
                         }, '')
                     );
                                                
                     field.onChange({original: files, derived: schema});
                 } catch (e) {
                     field.onChange({original: files, derived: e}); // If it'll never fail, the try/catch and the associated "validate" rule can be removed
                 }
             }}
         />
     }
/>
```

Sorry if it's all a bit confusing.. I actually don't know how to do this the best way, would probably need to find out through trial and error 🤷

DOBEN:
> Makes sense. With regards to the 2-step flow, it's possible to use browser location state, which can also be restored on back/forward browser navigation.
> 
> Anyway it would be nice if you could add an issue for this refactoring 🙂

https://github.com/Concordium/concordium-smart-contract-tools/issues/130

front-end-tools/src/reading_from_blockchain.ts
DOBEN: The advantage of validating the input parameters in the function is that we can simply have any thrown error displayed by setting `setTransactionErrorUpdate((err as Error).message)`. 

```
const tx = write(
    ...
 );

tx.then(setTxHashUpdate).catch((err: Error) =>
    setTransactionErrorUpdate((err as Error).message)
);
```


Validate user input/connection state prior to invocation of these functions would mean we need to add an extra `try{}.catch{}` blocks that has the same structure in `if {} else {}` statements as these functions (feels like we duplicate the  `if {} else {}` statements). I would suggest staying with the current way of evaluating the input/connection state in the functions. What do you think?

```
try {
   if (hasInputParameter) {
       if (inputParameter === undefined) {
          throw new Error(`Set input parameter`);
       }
    }
      ...
    } catch {
         setTransactionErrorUpdate((err as Error).message)
   }
}
```

front-end-tools/src/Main.tsx
DOBEN: The four steps/boxes have been moved into individual components and the `react-hook-form` library was applied for each form. The styling of the `react-hook-form` library seems much better in my opinion (good suggestion) (:+1: )

front-end-tools/src/components/DeployComponent.tsx
```diff
@@ -0,0 +1,268 @@
...
+                        }
+                    })
+                    .catch((e) => {
+                        setModuleReferenceDeployed(undefined);
+                        setViewErrorModuleReference((e as Error).message);
+                        clearInterval(interval);
+                    });
+            }, REFRESH_INTERVAL.asMilliseconds());
+        }
+    }, [connection, account, client, txHashDeploy]);
```
abizjak, line 89: Is it necessary to rerun this if `account` changes? This seems to only depend on the txHash, connection and client.

front-end-tools/src/components/DeployComponent.tsx
```diff
@@ -0,0 +1,268 @@
...
+    const [uploadError, setUploadError] = useState<string | undefined>(undefined);
+    const [isModuleReferenceAlreadyDeployedStep1, setIsModuleReferenceAlreadyDeployedStep1] = useState(false);
+    const [txHashDeploy, setTxHashDeploy] = useState<string | undefined>(undefined);
+    const [base64Module, setBase64Module] = useState<string | undefined>(undefined);
+    const [viewErrorModuleReference, setViewErrorModuleReference] = useState<string | undefined>(undefined);
+
+    // Refresh moduleReference periodically.
+    // eslint-disable-next-line consistent-return
+    useEffect(() => {
+        if (connection && client && account && txHashDeploy !== undefined) {
```
abizjak, line 64: Why is it necessary that the `account` is defined here?

front-end-tools/src/components/DeployComponent.tsx
```diff
@@ -0,0 +1,268 @@
...
+                        setIsModuleReferenceAlreadyDeployedStep1(false);
+                    } else {
+                        setIsModuleReferenceAlreadyDeployedStep1(true);
+                    }
+                })
+                .catch(() => {
+                    setIsModuleReferenceAlreadyDeployedStep1(false);
+                });
+        }
+    }, [connection, account, client, moduleReferenceCalculated]);
```
abizjak, line 106: Similarly here, is the `account` really necessary?

front-end-tools/src/components/DeployComponent.tsx
```diff
@@ -0,0 +1,268 @@
...
+                            const files = form.getValues('file');
+
+                            if (files !== undefined && files !== null && files.length > 0) {
+                                const file = files[0];
+                                const arrayBuffer = await file.arrayBuffer();
+
+                                const module = btoa(
+                                    new Uint8Array(arrayBuffer).reduce((data, byte) => {
+                                        return data + String.fromCharCode(byte);
+                                    }, '')
```
abizjak, line 149: I don't understand why we use `fromCharCode` here. If that's **the way** that's fine, but please add a comment since it's an odd construction.

front-end-tools/src/components/DeployComponent.tsx
```diff
@@ -0,0 +1,268 @@
...
+                                // Unversioned modules `.wasm` cannot be created by Concordium's tooling anymore.
+                                // If the module is versioned, the first 8 bytes are the version, followed by the `magicValue` below.
+                                // If the module is an old unversioned one, the module starts with the `magicValue` below.
+                                const magicValue = new Uint8Array([0x00, 0x61, 0x73, 0x6d]);
+                                let uploadedModuleFirst4Bytes = new Uint8Array([]);
+
+                                if (arrayBuffer.byteLength >= 4) {
+                                    uploadedModuleFirst4Bytes = new Uint8Array(arrayBuffer).subarray(0, 4);
+                                } else {
+                                    setUploadError(`You might have not uploaded a Concordium module.`);
```
abizjak, line 167: That's too weak. It is not a valid Wasm module. It needs to be at least 4 bytes.

front-end-tools/src/components/DeployComponent.tsx
```diff
@@ -0,0 +1,268 @@
...
+                                setBase64Module(module);
+                                setModuleReferenceCalculated(
+                                    Buffer.from(sha256([new Uint8Array(arrayBuffer)])).toString('hex')
+                                );
+
+                                // Concordium's tooling create versioned modules e.g. `.wasm.v1` now.
+                                // Unversioned modules `.wasm` cannot be created by Concordium's tooling anymore.
+                                // If the module is versioned, the first 8 bytes are the version, followed by the `magicValue` below.
+                                // If the module is an old unversioned one, the module starts with the `magicValue` below.
+                                const magicValue = new Uint8Array([0x00, 0x61, 0x73, 0x6d]);
```
abizjak, line 161: Please add that this is the magic value for Wasm modules specified by the Wasm spec.

front-end-tools/src/components/DeployComponent.tsx
```diff
@@ -0,0 +1,268 @@
...
+                                );
+
+                                setBase64Module(module);
+                                setModuleReferenceCalculated(
+                                    Buffer.from(sha256([new Uint8Array(arrayBuffer)])).toString('hex')
+                                );
+
+                                // Concordium's tooling create versioned modules e.g. `.wasm.v1` now.
+                                // Unversioned modules `.wasm` cannot be created by Concordium's tooling anymore.
+                                // If the module is versioned, the first 8 bytes are the version, followed by the `magicValue` below.
```
abizjak, line 159: The first 4 bytes are the version I think, the next 4 are the length, followed by the Wasm module.

front-end-tools/src/components/DeployComponent.tsx
```diff
@@ -0,0 +1,268 @@
...
+                                        }`
+                                    );
+                                }
+
+                                if (wasmModule) {
+                                    const moduleFunctions = WebAssembly.Module.exports(wasmModule);
+
+                                    const contractNames = [];
+                                    for (let i = 0; i < moduleFunctions.length; i += 1) {
+                                        if (moduleFunctions[i].name.slice(0, 5) === 'init_') {
```
abizjak, line 190: Isn't there a `startsWith` method?

front-end-tools/src/components/DeployComponent.tsx
```diff
@@ -0,0 +1,268 @@
...
+                                            contractNames.push(moduleFunctions[i].name.slice(5));
+                                        }
+                                    }
+                                    setContracts(contractNames);
+
+                                    const customSection = WebAssembly.Module.customSections(
+                                        wasmModule,
+                                        'concordium-schema'
```
abizjak, line 198: This is the current version. If somebody has an older version there might be `concordium-schema-v1` and `concordium-schema-v2` sections as well which have the "unversioned module schema".

There are probably some modules that are deployed still with the old schema.

front-end-tools/src/components/InitComponent.tsx
```diff
@@ -0,0 +1,643 @@
...
+    account: string;
+    connection: WalletConnection;
+    client: ConcordiumGRPCClient | undefined;
+    contracts: string[];
+    embeddedModuleSchemaBase64: undefined | string;
+    moduleReferenceCalculated: undefined | string;
+    moduleReferenceDeployed: undefined | string;
+}
+
+export default function InitComponenet(props: ConnectionProps) {
```
abizjak, line 35:
```suggestion
export default function InitComponent(props: ConnectionProps) {
```

abizjak: Also please remember to add a high-level comment to each of the components explaining its purpose.

front-end-tools/src/components/InitComponent.tsx
```diff
@@ -0,0 +1,643 @@
...
+    const [smartContractIndex, setSmartContractIndex] = useState<string | undefined>(undefined);
+    const [inputParameterTemplate, setInputParameterTemplate] = useState<string | undefined>(undefined);
+    const [displayContracts, setDisplayContracts] = useState<string[]>([]);
+
+    // Refresh smartContractIndex periodically.
+    // eslint-disable-next-line consistent-return
+    useEffect(() => {
+        if (connection && client && account && txHash !== undefined) {
+            const interval = setInterval(() => {
+                console.log('refreshing_smartContractIndex');
```
abizjak, line 94: I urge you to remove this logging as I mentioned above.

front-end-tools/src/components/InitComponent.tsx
```diff
@@ -0,0 +1,643 @@
...
+                        }
+                    })
+                    .catch((e) => {
+                        setSmartContractIndex(undefined);
+                        setSmartContractIndexError((e as Error).message);
+                        clearInterval(interval);
+                    });
+            }, REFRESH_INTERVAL.asMilliseconds());
+        }
+    }, [connection, account, client, txHash]);
```
abizjak, line 123: Same as for module deployment, I am not sure why the `account` is needed here (and above in the `if` statement. It is not used.

front-end-tools/src/components/InitComponent.tsx
```diff
@@ -0,0 +1,643 @@
...
+
+    useEffect(() => {
+        if (connection && client && account && moduleReference) {
+            if (moduleReference.length === 64) {
```
abizjak, line 127: Why check the length, but not the fact that it's a valid hex string?

front-end-tools/src/components/InitComponent.tsx
```diff
@@ -0,0 +1,643 @@
...
+                2
+            );
+
+            initTemplate = displayTypeSchemaTemplate(inputParamterTypeSchemaBuffer);
+
+            setInputParameterTemplate(initTemplate);
+        } catch (e) {
+            if (useModuleReferenceFromStep1) {
+                setSchemaError(
+                    `Could not get embedded schema from the uploaded module. Uncheck "Use Module from Step 1" checkbox to manually upload a schema or uncheck "Has Input Paramter" checkbox if this entrypoint has no input parameter.  Original error: ${e}`
```
abizjak, line 197: Would it not make sense to instead just offer to accept a raw parameter (in the form of a hex string)?

This allows somebody to use a pre-serialized string, or to deploy a new instance from parameters of a different init transaction without caring about the schema.

front-end-tools/src/components/InitComponent.tsx
```diff
@@ -0,0 +1,643 @@
...
+
+                            <Select
+                                options={displayContracts?.map((contract) => ({
+                                    value: contract,
+                                    label: contract,
+                                }))}
+                                onChange={(e) => {
+                                    form.setValue('smartContractName', e?.value);
+                                }}
+                            />
```
abizjak, line 363: I think it would  make sense to pre-select  the first one instead of nothing. In almost all cases there will be exactly one contract, so it's annoying to have to manually select it.

front-end-tools/src/components/InitComponent.tsx
```diff
@@ -0,0 +1,643 @@
...
+
+                                    setParsingError(undefined);
+                                }}
+                            />
+                            <Form.Text />
+                        </Form.Group>
+
+                        {(inputParameterType === 'number' || inputParameterType === 'string') && (
+                            <Form.Group className="mb-3">
+                                <Form.Label> Add your input parameter ({inputParameterType}):</Form.Label>
```
abizjak, line 532: It's a bit odd that this is here when the error is as well.

image

Whatever I put into that input field will be unusable. So it'd be better to not have it.

front-end-tools/src/components/ReadComponent.tsx
```diff
@@ -0,0 +1,525 @@
...
+
+        promise
+            .then((value) => {
+                setReturnValue(value);
+            })
+            .catch((err: Error) => setError((err as Error).message));
+    }
+
+    return (
+        <Box header="Read From Smart Contract">
```
abizjak, line 156: I think this is not really the whole story. You can **simulate** a contract call in this flow, not just getting back return values.

I think that should be emphasised as well.

front-end-tools/src/components/ReadComponent.tsx
```diff
@@ -0,0 +1,525 @@
...
+
+                <Button variant="primary" type="submit">
+                    Read Smart Contract
+                </Button>
+                <br />
+                <br />
+                {(deriveContractInfo ? embeddedModuleSchemaBase64 : uploadedModuleSchemaBase64) === undefined && (
+                    <Alert variant="warning">
+                        {' '}
+                        Warning: ModuleSchema is undefined. Return value might not be correctly decoded.{' '}
```
abizjak, line 506: Could you rephrase this into something more user-friendly?

Somethign like

```
There is no module schema, so the return value cannot be decoded.
```

front-end-tools/src/components/WriteComponent.tsx
```diff
@@ -0,0 +1,618 @@
...
+                            }
+                        }
+                    })
+                    .catch((e) => {
+                        setTransactionOutcome(`Fail; Error: ${(e as Error).message}`);
+                        clearInterval(interval);
+                    });
+            }, REFRESH_INTERVAL.asMilliseconds());
+        }
+    }, [connection, account, client, txHashUpdate]);
```
abizjak, line 109: Same comment about `accoutn`.

front-end-tools/src/components/WriteComponent.tsx
```diff
@@ -0,0 +1,618 @@
...
+            BigInt(data.maxExecutionEnergy),
+            BigInt(data.smartContractIndex),
+            data.cCDAmount ? BigInt(data.cCDAmount) : BigInt(0)
+        );
+
+        tx.then(setTxHashUpdate).catch((err: Error) => setTransactionErrorUpdate((err as Error).message));
+    }
+
+    return (
+        <Box header="Write To Smart Contract">
```
abizjak, line 200: It would be better to call this `update smart contract instance`. That would fit with the terminology we use elsewhere.

front-end-tools/src/components/WriteComponent.tsx
```diff
@@ -0,0 +1,618 @@
...
+                        </div>
+                        <br />
+                    </>
+                )}
+
+                <Form.Group className="mb-3 d-flex justify-content-center">
+                    <Form.Check
+                        type="checkbox"
+                        id="hasInputParameter"
+                        label="Has Input Parameter"
```
abizjak, line 409: When I click on the label here (not the checkbox) then the page scrolls to the top. That's odd.

(I discovered this by accident :))

front-end-tools/src/components/WriteComponent.tsx
```diff
@@ -0,0 +1,618 @@
...
+                                {...form.register('inputParameterType')}
+                                onChange={(e) => {
+                                    form.setValue('inputParameterType', e?.value);
+                                    form.setValue('inputParameter', undefined);
+
+                                    setParsingError(undefined);
+                                }}
+                            />
+                            <Form.Text />
+                        </Form.Group>
```
abizjak, line 496: I don't understand why I have to choose something here when it's clear what the parameter must be. For example

image


There the parameter is a string, that's what it says in the template. But I need to select the input parameter type as well. Why?

front-end-tools/src/constants.ts
```diff
@@ -19,3 +19,12 @@ export const EXAMPLE_JSON_OBJECT = {
 // These are the example arrays that are shown in the input parameter textarea as a placeholder when the user has no embedded schema in the module
 // or does not want to use the embedded schema (meaning if the checkbox "Use module from step 1" is unchecked).
 export const EXAMPLE_ARRAYS = 'Examples: \n\n[1,2,3] or \n\n["abc","def"] or \n\n[{"myFieldKey":"myFieldValue"}]';
+
+export const INPUT_PARAMETER_TYPES_OPTIONS = [
```
abizjak, line 23: Please document.

front-end-tools/src/reading_from_blockchain.ts
```diff
@@ -0,0 +1,146 @@
...
+        throw new Error(`rpcClient undefined`);
+    }
+    if (moduleRef === undefined) {
+        throw new Error(`Set module ref`);
+    }
+
+    return rpcClient.getEmbeddedSchema(moduleRef);
+}
+
+export async function read(
```
abizjak, line 45: Document please.

front-end-tools/src/reading_from_blockchain.ts
```diff
@@ -0,0 +1,146 @@
...
+    rpcClient: ConcordiumGRPCClient | undefined,
+    contractName: string | undefined,
```
abizjak, line 47: Why does this need to take parameters that are `| undefined`?

front-end-tools/src/reading_from_blockchain.ts
```diff
@@ -0,0 +1,146 @@
...
+        toBuffer(res.returnValue, 'hex'),
+        toBuffer(moduleSchema, 'base64'),
+        contractName,
+        entryPoint
+    );
+
+    if (returnValue === undefined) {
+        throw new Error(
+            `Deserializing the returnValue from the '${contractName}.${entryPoint}' method of contract '${contractIndex}' failed`
+        );
```
abizjak, line 142: Don't you get any useful error message here?

front-end-tools/src/writing_to_blockchain.ts
```diff
@@ -20,97 +21,183 @@ export async function deploy(connection: WalletConnection, account: string, base
     } as DeployModulePayload);
 }
 
+export async function write(
+    connection: WalletConnection | undefined,
+    account: string | undefined,
+    inputParameter: string | undefined,
+    contractName: string | undefined,
```
abizjak, line 28: Same comments as for `read`.

Not sure why all these `|undefined` are needed in this function. I'd rather if they are handled in the form validation already.

front-end-tools/src/components/ReadComponent.tsx
```diff
@@ -0,0 +1,525 @@
...
+                                            const target = event.target as HTMLTextAreaElement;
+
+                                            try {
+                                                JSON.parse(target.value);
+                                            } catch (e) {
+                                                setParsingError((e as Error).message);
+                                                return;
+                                            }
+                                            form.setValue('inputParameter', target.value);
+                                        }}
```
soerenbf, line 457: Refactor to validate function

front-end-tools/src/components/ReadComponent.tsx
```diff
@@ -0,0 +1,525 @@
...
```
soerenbf, line 476: Same here

front-end-tools/src/components/WriteComponent.tsx
```diff
@@ -0,0 +1,618 @@
...
+    };
+
+    const form = useForm<FormType>();
+    const deriveContractInfo = form.watch('deriveFromSmartContractIndex');
+    const hasInputParameter = form.watch('hasInputParameter');
+    const entryPointName = form.watch('entryPointName');
+    const smartContractName = form.watch('smartContractName');
+    const inputParameterType = form.watch('inputParameterType');
+
+    const isPayable = form.watch('isPayable');
```
soerenbf, line 56: Would be nice to refactor this to [`useWatch`](https://react-hook-form.com/docs/usewatch) as well, as it is recommended to use to trigger side effects.

front-end-tools/src/components/WriteComponent.tsx
```diff
@@ -0,0 +1,618 @@
...
+                                    clearInterval(interval);
+                                }
+                            }
+                        }
+                    })
+                    .catch((e) => {
+                        setTransactionOutcome(`Fail; Error: ${(e as Error).message}`);
+                        clearInterval(interval);
+                    });
+            }, REFRESH_INTERVAL.asMilliseconds());
```
soerenbf, line 107: Clear interval when deps change?

front-end-tools/src/components/WriteComponent.tsx
```diff
@@ -0,0 +1,618 @@
...
+        }
+    }, [connection, account, client, txHashUpdate]);
+
+    useEffect(() => {
+        if (entryPointTemplate !== undefined && hasInputParameter === false) {
+            setShouldWarnInputParameterInSchemaIgnored(true);
+        } else {
+            setShouldWarnInputParameterInSchemaIgnored(false);
+        }
+    }, [entryPointTemplate, hasInputParameter]);
```
soerenbf, line 117: Refactor to `useMemo`

front-end-tools/src/components/WriteComponent.tsx
```diff
@@ -0,0 +1,618 @@
...
+                                            const target = event.target as HTMLTextAreaElement;
+
+                                            try {
+                                                JSON.parse(target.value);
+                                            } catch (e) {
+                                                setParsingError((e as Error).message);
+                                                return;
+                                            }
+                                            form.setValue('inputParameter', target.value);
+                                        }}
```
soerenbf, line 539: Refactor to validate function

front-end-tools/src/components/WriteComponent.tsx
```diff
@@ -0,0 +1,618 @@
...
```
soerenbf, line 558: And here

soerenbf (commented):
I still think it's worthwhile to look into the structure of these inputs that trigger other changes (effects or derived values), which is my main issue with the code as it is right now.

I would prefer if we stuck with the recommendations from the documentation of the forms library, which revolve around

- spread `form.register` onto input
- use `Controller` to integrate with controlled field components
- use validation rules and associated error messages registered
- use `useWatch` to handle complicated derived values or effects

front-end-tools/src/components/DeployComponent.tsx
DOBEN: removed

front-end-tools/src/components/InitComponent.tsx
DOBEN: removed

front-end-tools/src/components/InitComponent.tsx
DOBEN: I agree with you that as long as we do not have a valid schema, the front end will not be able to parse the input parameter and create a valid transaction (with that input parameter). 

On the other hand, it is not good to wipe out the input parameter field too often (e.g. hiding it every time an error occurs). 
Imagine you have a complex object inputted as your input parameter there and when typing your 'entrypoint' name you create a typo, suddenly an error occurs and even though you could easily just correct the typo in the `entrypoint` name field, you have to start again to input your input parameter object in this field.

On the other hand, it allows us to display the example objects/array.

front-end-tools/src/components/ReadComponent.tsx
DOBEN: Currently, this box is set up to actually read a value from the smart contract and display the `return_value` to the user (either the raw bytes or deserialized if a schema is available). No dry run/simulation can be done with the box.

Also, the `invoke` call can be used to dry-run an `update` transaction, such transactions usually have no `return_value`, so I wouldn't want to treat them in this box. Maybe a separate `simulation` box would be more appropriate.

If I recall correctly we intended to add `dry-run` steps before the deploy/init/update transactions at some future steps which would be the best option in my opinion.
 
 I start collection feature requests here:
 https://github.com/Concordium/concordium-smart-contract-tools/issues/131

front-end-tools/src/reading_from_blockchain.ts
DOBEN: Removed the `| undefined` at `contractName`.

Many other parameters return type errors if you remove the `|undefined`. E.g. the `entrypoint` parameter requires it because of this line:
```
form.setValue('entryPointName', e?.value);
```

front-end-tools/src/writing_to_blockchain.ts
DOBEN: see comment above.

front-end-tools/src/components/WriteComponent.tsx
DOBEN: How do you else want to do this programmatically? We have several templates but I am not aware of a good/global mapping between the templates and types. Hardcoding a long mapping here at the front end that I just defined seems also weird.

front-end-tools/src/reading_from_blockchain.ts
DOBEN: I return the original response object/error as well now.

DOBEN: If the function returns no `return_value`, it is an error for me (miss use of this box). Dry running should be implemented before executing the `init/deploy/update` transaction in the other boxes instead.

DOBEN: The error is now returned as well from the `deserializeReceiveReturnValue` function.
```

    let returnValue;

    try {
        // If schema is provided deserialize return value
        returnValue = deserializeReceiveReturnValue(
            toBuffer(res.returnValue, 'hex'),
            toBuffer(moduleSchema, 'base64'),
            contractName,
            entryPoint
        )
    } catch (e) {
        throw new Error(
            `Deserializing the returnValue from the '${contractName}.${entryPoint}' method of contract '${contractIndex}' failed. Original: ${e}`
        );
    }

    if (returnValue === undefined) {
        throw new Error(
            `Deserializing the returnValue from the '${contractName}.${entryPoint}' method of contract '${contractIndex}' failed.`
        );
    } else {
        return JSON.stringify(returnValue);
    }
```

front-end-tools/src/components/DeployComponent.tsx
DOBEN: I recall that initially, I was converting the whole file in one go but that didn't work for our largest smart contracts. The `reduce` construct enables the conversion in smaller junks and the error went away.

front-end-tools/src/components/DeployComponent.tsx
DOBEN: True, but since Concordium tooling does not produce such old modules anymore and we assume that the `DeployComponent` is used by current smart contract developers who are actively developing and compiling fresh modules, it might be ok to not support very old modules here.

If you feel strongly about it, I suggest we add this as a future task.

front-end-tools/src/components/InitComponent.tsx
DOBEN: This is very useful for experienced developers also in the context of e.g. using proxy-implementation contracts where the schema of a contract might not include all the options on how to interact with the contract.

Since this feature needs some substantial changes, I added it to the top of the requested feature list here:
https://github.com/Concordium/concordium-smart-contract-tools/issues/131

front-end-tools/src/components/WriteComponent.tsx
DOBEN: I think this is caused because some of the `id`s are duplicated. Meaning if you clicked on the field it jumped to the first checkbox with this id on the page (which was somewhere at the top). I removed duplicated `id`s.

front-end-tools/src/Main.tsx
```diff
@@ -217,172 +64,17 @@ export default function Main(props: ConnectionProps) {
...
-            const element = inputParameterTextAreaRef.current as unknown as HTMLSelectElement;
-            element.value = getArrayExample(template);
-        } else if (dropDown === 'object') {
-            const element = inputParameterTextAreaRef.current as unknown as HTMLSelectElement;
-            element.value = getObjectExample(template);
-        }
-    }, [hasInputParameter, useModuleFromStep1, contractName, uploadedModuleSchemaBase64, dropDown]);
-
     useEffect(() => {
         if (connection && client && account) {
```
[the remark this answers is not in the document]
DOBEN: This bottom `useEffect` triggers when the page is loaded to immediately display the account balance. While the top `useEffect` starts after the first `REFRESH_INTERVAL` seconds have passed to then periodically poll for any updates in the account balance.

Without the bottom `useEffect`, you would have to wait until the first `REFRESH_INTERVAL` seconds have passed before seeing any account balance info.

front-end-tools/src/Main.tsx
```diff
@@ -217,172 +64,17 @@ export default function Main(props: ConnectionProps) {
...
+                        setViewErrorAccountInfo(undefined);
                     })
                     .catch((e) => {
-                        setAccountBalance('');
+                        setAccountBalance(undefined);
                         setViewErrorAccountInfo((e as Error).message.replaceAll('%20', ' '));
                         setAccountExistsOnNetwork(false);
                     });
             }, REFRESH_INTERVAL.asMilliseconds());
-            return () => clearInterval(interval);
```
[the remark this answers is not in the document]
DOBEN: yes, you are right. I moved that back.

front-end-tools/src/components/InitComponent.tsx
```diff
@@ -0,0 +1,643 @@
...
+
+                                register.onChange(e);
+
+                                setModuleReferenceLengthError(undefined);
+
+                                const moduleRef = form.getValues('moduleReference');
+
+                                if (moduleRef !== undefined && moduleRef.length !== 64) {
+                                    setModuleReferenceLengthError('Module reference has to be of length 64');
+                                }
```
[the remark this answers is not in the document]
DOBEN: I experimented with it but the problem is that we want to give feedback to the user as early as possible for some of the critical input fields (input values that are needed for e.g. loading additional information such as schemas). That is why some custom evaluations and more precise error reporting are done here to guide the user to resolve the issue.

If we wait until the user presses the `submitForm` button or give a very general error here (input value does not match defined pattern) seems not so helpful to the user.

front-end-tools/src/components/InitComponent.tsx
```diff
@@ -0,0 +1,643 @@
...
+                                            const target = event.target as HTMLTextAreaElement;
+
+                                            try {
+                                                JSON.parse(target.value);
+                                            } catch (e) {
+                                                setParsingError((e as Error).message);
+                                                return;
+                                            }
+                                            form.setValue('inputParameter', target.value);
+                                        }}
```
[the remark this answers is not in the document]
DOBEN: I experimented with it but the problem is that we want to give feedback to the user as early as possible for some of the critical input fields (input values that are needed for e.g. loading additional information such as schemas). That is why some custom evaluations and more precise error reporting are done here to guide the user to resolve the issue.

If we wait until the user presses the submitForm button or give a very general error here (input value does not match defined pattern) seems not so helpful to the user.

front-end-tools/src/components/InitComponent.tsx
```diff
@@ -0,0 +1,643 @@
...
```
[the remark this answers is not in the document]
DOBEN: I think the correct way to avoid duplication would be to add additional components since this is not the only part that is `duplicated`. I added it as a future task and keep it as it is for now.

```
Refactor: The text area used as input parameter should be its own component to avoid code duplication. The `inputParameterType` dropDown and the `inputParameter` field/textArea should be its own component to avoid code duplication.
```
https://github.com/Concordium/concordium-smart-contract-tools/issues/131

front-end-tools/src/components/DeployComponent.tsx
```diff
@@ -0,0 +1,268 @@
...
+                    <Form.Control
+                        type="file"
+                        accept=".wasm,.wasm.v0,.wasm.v1"
+                        {...form.register('file')}
+                        onChange={async (e) => {
+                            const register = form.register('file');
+
+                            register.onChange(e);
+
+                            setUploadError(undefined);
```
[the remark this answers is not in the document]
DOBEN: I added it as a future task:
```
Refactor: Investigate `Controller` component and `Validation rules` for the purpose of reducing the `onChange` functions.```

File: front-end-tools/src/components/InitComponent.tsx [hunk 48]
```diff
@@ -0,0 +1,643 @@
...
+                <Form.Group className="mb-3 d-flex justify-content-center">
+                    <Form.Check
+                        type="checkbox"
+                        id="useModuleReferenceFromStep1"
+                        label="Use Module From Step 1"
+                        {...form.register('useModuleReferenceFromStep1')}
+                        onChange={async (e) => {
+                            const register = form.register('useModuleReferenceFromStep1');
+
+                            register.onChange(e);
```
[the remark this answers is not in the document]
DOBEN: Refactor: Investigate `Controller` component and `Validation rules` for the purpose of reducing the `onChange` functions.

front-end-tools/src/components/ReadComponent.tsx
```diff
@@ -0,0 +1,525 @@
...
+        inputParameterType: string | undefined;
+        inputParameter: string | undefined;
+    };
+
+    const form = useForm<FormType>();
+    const deriveContractInfo = form.watch('deriveFromSmartContractIndex');
+    const hasInputParameter = form.watch('hasInputParameter');
+    const entryPointName = form.watch('entryPointName');
+    const smartContractName = form.watch('smartContractName');
+    const inputParameterType = form.watch('inputParameterType');
```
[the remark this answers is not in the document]
DOBEN: It seems to need a controller as well, so would leave it for future investigation/tasks to investigate if we want to introduce controllers or not.

front-end-tools/src/components/InitComponent.tsx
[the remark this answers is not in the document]
DOBEN: The module `89d2d62a5a4f7a88c7689fff9dcf783469f4ad7139f9f0f71d8f53abe6dc5a7c` has two smart contracts and you can select the smart contractName via the dropDown without problems.

front-end-tools/src/components/InitComponent.tsx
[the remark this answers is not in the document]
DOBEN: see below comment

front-end-tools/src/Main.tsx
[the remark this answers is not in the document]
soerenbf: Yes but what I was referring to is that if the body of the interval callback and the initial load is the same, it would make sense to have make them as one function used twice

front-end-tools/src/components/DeployComponent.tsx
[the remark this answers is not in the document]
soerenbf: Alright, if we need to get this out I guess that will have to do 😃

front-end-tools/src/components/InitComponent.tsx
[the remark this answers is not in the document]
soerenbf: It might make sense to validate instantly without waiting for submission for these forms, but the library also supports that: https://react-hook-form.com/docs/useform#mode

```ts
const form = useForm({mode: 'all'});
```

front-end-tools/src/components/InitComponent.tsx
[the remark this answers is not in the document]
soerenbf: Ah sorry, I can see how that is easy to misunderstand. What I mean is that you don't call `form.register('smartContractName')` in the case where the dropdown is set, which is why I referred to the `Controller` example from stackoverflow.

front-end-tools/src/components/InitComponent.tsx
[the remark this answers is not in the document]
soerenbf: Okay. But I would still argue that this should be refactored to use the forms library as intended.

front-end-tools/src/components/ReadComponent.tsx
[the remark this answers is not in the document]
soerenbf: I doesn't depend on using `Controller` components, just the `control` object provided by `useForm`, which is just a way to hook the components and functions provided by the library up to the form.

I don't understand your restraint from using this functionality? It's not something that makes anything behave differently. If you want to leave it for later, fair enough, however I fear this will become something that lingers in the backlog.

soerenbf (commented):
Tested the different parts of the app again, and experienced this discrepancy:

Screenshot 2023-11-09 at 09 40 38

If you want the concrete module deployed, I can send it to you.

Other than that, I don't have any more comments if all the optimizations with regards to using the forms library are pushed to sometime in the future.

front-end-tools/src/components/InitComponent.tsx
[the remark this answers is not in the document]
DOBEN: I added the `{mode: 'all'}` to all forms.

front-end-tools/src/components/InitComponent.tsx
```diff
@@ -0,0 +1,641 @@
...
+                    </>
+                )}
+
+                {moduleReferenceError && <Alert variant="danger"> Error: {moduleReferenceError}. </Alert>}
+
+                <Row>
+                    <Form.Group className="col-md-4 mb-3">
+                        <Form.Label> Module Reference</Form.Label>
+                        <Form.Control
+                            defaultValue="91225f9538ac2903466cc4ab07b6eb607a2cd349549f357dfdf4e6042dde0693"
```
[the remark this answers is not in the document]
DOBEN: hmm, that would suggest that whenever we upload a new module in step 1 we automatically update values in box 2 (such as this `defaultValue` as well as its underlying field value). I am a bit worried that this will get confusing at some point if values across boxes change so I tried to avoid it.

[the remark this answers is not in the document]
abizjak: It's fine to not do this now, but it's not the best experience as it is now either.

Perhaps this will be cleaner when/if the pages are split into the different flows, so that not everything is on the same page.

front-end-tools/src/components/DeployComponent.tsx
DOBEN: The `Update` and `Read` step get their embedded module from `rpcClient.getEmbeddedSchema(moduleRef)` which should support old schemas as long as we haven't removed support for it.

front-end-tools/src/components/ReadComponent.tsx
DOBEN: Added it to:
https://github.com/Concordium/concordium-smart-contract-tools/issues/131
```
Better error handling. In particular, think about using the read box as a simulation/dry-run box.
https://github.com/Concordium/concordium-smart-contract-tools/pull/123#discussion_r1387772430
```

Issue #138, opened 2023-11-23
VSCode command for generating smart contract clients


Bargsteen:
**Task description**

Add a VSCode command for generating smart contract clients via our ccd-js-gen tool.

PR #140, opened 2023-11-28
Support generating TypeScript/JavaScript clients using ccd-js-gen

Bargsteen:
## Purpose

Integrates our ccd-js-gen tool in our VS Code extension, so that js/ts clients can be generated from smart contract modules.
Closes https://github.com/Concordium/concordium-smart-contract-tools/issues/138.

As discussed in standup, I could not use ccd-js-gen as a library because it is an ES module, which VS Code does not support. ES modules are supported in Electron 28+ (https://github.com/electron/electron/pull/37535), but VS Code is still on v25.

We tested the new commands on Windows, and the generator command did not work. After some debugging, we found out that it is because of a badly generated `ps1`/`cmd` file that is used as the executable.
Should be fixed by https://github.com/Concordium/concordium-node-sdk-js/pull/315, but I have not tested the complete fix on Windows.

## Changes

- Add commands:
  - To get the version of the ccd-js-gen executable
  - To generate js/ts clients from a wasm module
- Add a task for generating js/ts clients
- Add settings
  - Custom ccd-js-gen executable
  - Additional arguments for ccd-js-gen
- Refactorings
   - Move task types to a new file `tasks.ts` as they are needed for both ccd-js-gen and cargo concordium
   - Move helper functions for calling executables into the file `executable-wrapper-helpers.ts` 

## Checklist

      hard-to-understand areas.


abizjak (commented):
Looks good, but I do not understand how the ccd-gen-js is embedded as a binary.

Also, what runtime will the ccd-js-gen run with?

vscode-smart-contracts/src/commands.ts
```diff
@@ -304,3 +319,59 @@ async function installCargoGenerate() {
...
+    if (action !== build_contract_action) {
+      return;
+    }
+    const task = await buildTask("outDir");
+    if (task === undefined) {
+      return;
+    }
+    const exitCode = await executeAndAwaitTask(task);
+    if (exitCode !== 0) {
+      return;
```
limemloh, line 360: **Question:** Is the user shown anything in this case, and will they understand why it did not generate anything in this case?

vscode-smart-contracts/src/commands.ts
```diff
@@ -304,3 +319,59 @@ async function installCargoGenerate() {
...
 }
+
+/**
+ * Generate JavaScript/TypeScript clients for the smart contract module.
+ */
+export async function generateJsClients() {
+  const editor = vscode.window.activeTextEditor;
+  if (editor === undefined) {
+    vscode.window.showErrorMessage(
+      "Unable to determine smart contract project. Open a file in the smart contract project that you want to build"
```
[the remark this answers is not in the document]
Bargsteen: Good catch ;)

vscode-smart-contracts/src/commands.ts
Bargsteen: Good catch. I am now throwing errors there instead of silently failing.
The case I had tested with was when it couldn't determine which project to build, in which case the build task threw an error on its own.

Issue #135, opened 2023-11-15
Deploy and Initialize SC Tools - Typos in error message


mh-concordium:
**Bug Description**
Some minor typos in the error message when trying to read/write to a contract without embedded schema.

**Steps to Reproduce**
Try to read or write to a contract without embedded schema. Don't upload any schema files.

**Expected Result**
I am not sure if we should just fix the error message like this:
`Error: Could not get schema from uploaded file. Uncheck "Has Input Parameter" checkbox if this entry point has no input parameter. Original error: Error: Unable to get parameter schema, due to: Missing Schema Version.`

Or if we should change the error message to something like  `Schema was not uploaded` but then I guess you can't us it as a generic message for all the errors.

**Actual Result**
`Error: Could not get schema from uploaded schema. Uncheck "Has Input Paramter" checkbox if this entrypoint has no input parameter. Original error: Error: Unable to get parameter schema, due to: Missing Schema Version.`

**Versions**
- Software Version: 3.0.1

PR #142, opened 2023-11-29
Fix front end tools error message

lassemoldrup:
## Purpose

Addresses #135.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/front-end-tools/src/components/ReadComponent.tsx
+++ b/front-end-tools/src/components/ReadComponent.tsx
@@ -96,14 +96,12 @@ export default function ReadComponenet(props: ConnectionProps) {
                 throw new Error('Set smart contract name');
             }
 
-            let schema = '';
-
-            const schemaFromModule = deriveContractInfo ? embeddedModuleSchemaBase64 : uploadedModuleSchemaBase64;
-
-            if (schemaFromModule !== undefined) {
-                schema = schemaFromModule;
+            const schema = deriveContractInfo ? embeddedModuleSchemaBase64 : uploadedModuleSchemaBase64;
+            if (schema === undefined) {
+                setSchemaError('Schema was not uploaded');
+                return;
             }
-
+            
             const readFunctionTemplate = getUpdateContractParameterSchema(
                 toBuffer(schema, 'base64'),
                 ContractName.fromString(smartContractName),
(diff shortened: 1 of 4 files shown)


DOBEN:
You can run `yarn lint --fix` before pushing the code to have your code linted (avoids the failing CI pipeline).

front-end-tools/src/components/ReadComponent.tsx
```diff
@@ -96,14 +96,12 @@ export default function ReadComponenet(props: ConnectionProps) {
...
 
-            let schema = '';
-
-            const schemaFromModule = deriveContractInfo ? embeddedModuleSchemaBase64 : uploadedModuleSchemaBase64;
-
-            if (schemaFromModule !== undefined) {
-                schema = schemaFromModule;
+            const schema = deriveContractInfo ? embeddedModuleSchemaBase64 : uploadedModuleSchemaBase64;
+            if (schema === undefined) {
+                setSchemaError('Schema was not uploaded');
```
DOBEN, line 101: If a first-time user does the following thing in the read box:
- inserts the contract `6956 (no embedded schema)`
- checks the checkbox `derive from smart contract index`
- selects an entry point name
- checks the checkbox `has input parameter`

Now the error occurs: `Schema was not uploaded`
Since there is nothing in the read box where the user could upload the schema, it is confusing.

Suggestion: Use the following error message instead which has additional information on how to resolve the above scenario.

`Schema was not uploaded or not embedded into the module. Uncheck "Derive From Smart Contract Index" checkbox to manually upload a schema or uncheck "Has Input Paramter" checkbox if this entrypoint has no input parameter. `

lassemoldrup: It seems to me that the "has input parameter" box shouldn't be there when deriving from the contract, but I don't know how difficult it would be to change.

lassemoldrup: Good catch!

front-end-tools/src/components/UpdateComponent.tsx
```diff
@@ -141,12 +141,10 @@ export default function UpdateComponenet(props: ConnectionProps) {
...
 
-            let schema = '';
-
-            const schemaFromModule = deriveContractInfo ? embeddedModuleSchemaBase64 : uploadedModuleSchemaBase64;
-
-            if (schemaFromModule !== undefined) {
-                schema = schemaFromModule;
+            const schema = deriveContractInfo ? embeddedModuleSchemaBase64 : uploadedModuleSchemaBase64;
+            if (schema === undefined) {
+                setSchemaError('Schema was not uploaded');
```
DOBEN, line 146: the same here please

DOBEN (approved):

PR #143, opened 2023-11-29
Fix sc frontend read box

### LINKED ISSUE #139: [Front-end-tools] Read Box bugs
**Bug Description**
Using the readBox (third box from the top):

- Issue 1: Read functions with no input parameter return `Error: Do not know how to serialize a BigInt.`
E.g. `DECIMALS` and `requiredSignatures` in contract instance 7283

- Issue 2: The `[object Object]` in the following error message should be fixed which appears when calling an entry point with e.g. the wrong input parameter.

```
Error: RPC call 'invokeContract' on method '[object Object].[object Object]' of contract '7283' failed. Reject reason: -2147483646.
```

**Steps to Reproduce**
-  Issue 1: Use the readBox. Try to invoke `DECIMALS` and `requiredSignatures` in contract instance 7283.
-  Issue 2: Use the readBox. Invoke an entry point with a wrong input parameter.

**Expected Result**
-  Issue 1: Should work. No error message.
-  Issue 2: No  '[object Object].[object Object]'  in error message.

**Actual Result**
-  Issue 1: `Error: Do not know how to serialize a BigInt.` is returned.
-  Issue 2: '[object Object].[object Object]'  in error message.

lassemoldrup:
## Purpose

Closes #139.

## Changes

- RPC calls could return BigInts, which cannot be serialized with `JSON.stringify`. This PR uses the package `json-bigint` as a replacement in places where a return value is serialized to address this.
- Contract method names are fixed in error messages.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/front-end-tools/CHANGELOG.md
+++ b/front-end-tools/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased changes
 
 - Fix error message when specifying input parameter without uploading schema.
+- Fix error when contract calls return BigInts.
 - Fix contract method names in error messages.
 
 ## 3.0.1
--- a/front-end-tools/CHANGELOG.md
+++ b/front-end-tools/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased changes
 
 - Fix error message when specifying input parameter without uploading schema.
+- Fix contract method names in error messages.
 
 ## 3.0.1
 
--- a/front-end-tools/CHANGELOG.md
+++ b/front-end-tools/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Fix error message when specifying input parameter without uploading schema.
 - Fix error when contract calls return BigInts.
 - Fix contract method names in error messages.
+- Give warning when deploying a module that does not have embedded build information.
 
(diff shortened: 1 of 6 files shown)


front-end-tools/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Unreleased changes
 
 - Fix error message when specifying input parameter without uploading schema.
+- Fix contract method names in error messages.
```
DOBEN, line 4: Please also mention that you fixed the bigint issue.

DOBEN (approved):
Well done.

You might want to close the issue `Closes https://github.com/Concordium/concordium-smart-contract-tools/issues/139.` instead of `Addresses https://github.com/Concordium/concordium-smart-contract-tools/issues/139.`

DOBEN:
Suggestion: If we have no schema uploaded (and the read box returns the raw bytes), it would be better to return them e.g. in a hex-decimal type of string then the Buffer presentation (since for large objects it is hard to copy/read).

Screenshot from 2023-11-29 13-24-43

abizjak (approved):

PR #154, opened 2024-01-23
Add option to `deriveFromChain` given a module reference when initializing a contract instance

### LINKED ISSUE #149: Deploy and Initialize Smart Contracts Tool - enable initializing new smart contract from the existing module reference
**Description**
Currently it is not possible to initialize new smart contract instance without schema.bin file in case the module was not deployed in step 1. Therefore, it is not possible to just use any modules that exist on chain but rather only modules deployed in step 1.

Sctools application shows error as the schema was not embedded while it was.

**Steps to Reproduce**

1. On Testnet try to use `243bf19c23473ae4f0feea2f40fcfb6181a0b5a0b96baeb1b2fe986b27ffffaf` as module reference.
2. Smart contract name is election.
3. Check has input parameters.

**Expected Result**
It would be nice to be able to just use any module reference with schema from chain.

**Actual Result**
`Error: Set schema.`

**Versions**
- Software Version:3.0.1

DOBEN:
## Purpose

closes #149 
Until now, people had to upload a module file at step 1 before they were able to derive values for initializing in step 2.
Now, there is a new option to derive the values in step 2 from the chain given a valid module reference.

## Changes

- Simplified interactions between Step 1 and Step 2. Only the moduleRefrence is now a shared state between the two boxes.
- Add option to `deriveFromChain` given a module reference in Step 2.

### DIFF
--- a/front-end-tools/src/components/DeployComponent.tsx
+++ b/front-end-tools/src/components/DeployComponent.tsx
@@ -25,9 +25,6 @@ interface ConnectionProps {
     connection: WalletConnection;
     client: ConcordiumGRPCClient | undefined;
     isTestnet: boolean;
-    setContracts: (contracts: string[]) => void;
-    setEmbeddedModuleSchemaBase64Init: (embeddedModuleSchemaBase64Init: string) => void;
-    setModuleReferenceDeployed: (moduleReferenceDeployed: ModuleReference.Type | undefined) => void;
     setModuleReferenceCalculated: (moduleReferenceCalculated: ModuleReference.Type) => void;
     moduleReferenceCalculated: ModuleReference.Type | undefined;
 }
@@ -37,17 +34,7 @@ interface ConnectionProps {
  *  This components creates a `DeployModule` transaction.
  */
 export default function DeployComponenet(props: ConnectionProps) {
-    const {
-        isTestnet,
-        client,
-        connection,
-        account,
-        setContracts,
-        setModuleReferenceDeployed,
-        setModuleReferenceCalculated,
-        moduleReferenceCalculated,
-        setEmbeddedModuleSchemaBase64Init,
-    } = props;
+    const { isTestnet, client, connection, account, setModuleReferenceCalculated, moduleReferenceCalculated } = props;
 
     type FormType = {
         file: FileList | undefined;
@@ -76,9 +63,6 @@ export default function DeployComponenet(props: ConnectionProps) {
                                 report.outcome.summary.transactionType === TransactionKindString.DeployModule
                             ) {
                                 setTransactionOutcome('Success');
-                                setModuleReferenceDeployed(
-                                    ModuleReference.fromHexString(report.outcome.summary.moduleDeployed.contents)
-                                );
                                 clearInterval(interval);
                             } else {
                                 setTransactionOutcome('Fail');
@@ -87,7 +71,6 @@ export default function DeployComponenet(props: ConnectionProps) {
                         }
                     })
                     .catch((e) => {
-                        setModuleReferenceDeployed(undefined);
                         setTransactionOutcome(`Fail; Error: ${(e as Error).message}`);
                         clearInterval(interval);
                     });
@@ -122,7 +105,6 @@ export default function DeployComponenet(props: ConnectionProps) {
 
         const tx = deploy(connection, AccountAddress.fromBase58(account), base64Module);
         tx.then((txHash) => {
-            setModuleReferenceDeployed(undefined);
             setTxHashDeploy(txHash);
         }).catch((err: Error) => setTransactionErrorDeploy((err as Error).message));
     }
@@ -142,7 +124,6 @@ export default function DeployComponenet(props: ConnectionProps) {
                             register.onChange(e);
 
                             setUploadError(undefined);
-                            setModuleReferenceDeployed(undefined);
                             setTransactionErrorDeploy(undefined);
                             setTxHashDeploy(undefined);
 
@@ -197,33 +178,6 @@ export default function DeployComponenet(props: ConnectionProps) {
                                 }
 
                                 if (wasmModule) {
-                                    const moduleFunctions = WebAssembly.Module.exports(wasmModule);
-
-                                    const contractNames = [];
-                                    for (let i = 0; i < moduleFunctions.length; i += 1) {
-                                        if (moduleFunctions[i].name.startsWith('init_')) {
-                                            contractNames.push(moduleFunctions[i].name.slice(5));
-                                        }
-                                    }
-                                    setContracts(contractNames);
-
-                                    const customSection = WebAssembly.Module.customSections(
-                                        wasmModule,
-                                        'concordium-schema'
-                                    );
-
-                                    const schema = new Uint8Array(customSection[0]);
-
-                                    // Use `reduce` to be able to convert large schema.
-                                    const moduleSchemaBase64Embedded = btoa(
-                                        new Uint8Array(schema).reduce(
-                                            (data, byte) => data + String.fromCharCode(byte),
-                                            ''
-                                        )
-                                    );
-
-                                    setEmbeddedModuleSchemaBase64Init(moduleSchemaBase64Embedded);
-
                                     // Check if the module was built as a reproducible build.
                                     const buildInfoSection = WebAssembly.Module.customSections(
                                         wasmModule,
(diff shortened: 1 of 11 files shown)


front-end-tools/src/components/InitComponent.tsx
```diff
@@ -189,71 +193,107 @@ export default function InitComponent(props: ConnectionProps) {
...
+            if (!hexRegex.test(value)) {
+                return 'Invalid module reference. Not a hex string.';
             }
+
+            setModuleReference(ModuleReference.fromHexString(value));
+
+            return true;
+        } catch (e) {
+            return `Invalid module reference. Original error: ${(e as Error).message}`;
+        }
```
abizjak, line 243: Since the regex now does not ensure that the string is 64 characters, won't this also accept something like `aa`?

Is there additional validation on submission?

front-end-tools/src/components/InitComponent.tsx
```diff
@@ -189,71 +193,107 @@ export default function InitComponent(props: ConnectionProps) {
...
+
+            const contractNames = [];
+            for (let i = 0; i < moduleFunctions.length; i += 1) {
+                if (moduleFunctions[i].name.startsWith('init_')) {
+                    contractNames.push(moduleFunctions[i].name.slice(5));
+                }
+            }
+
+            setDisplayContracts(contractNames);
+            form.setValue('smartContractName', contractNames[0]);
```
abizjak, line 267: Could you document why this choice is made here?

front-end-tools/src/components/InitComponent.tsx
```diff
@@ -265,107 +305,125 @@ export default function InitComponent(props: ConnectionProps) {
...
-                                    setModuleReferenceLengthError(
-                                        'Module reference has to be a valid hex string `[0-9A-Fa-f]` of length 64'
-                                    );
-                                }
-                            }}
+                            defaultValue={MODULE_REFERENCE_PLACEHOLDER}
+                            disabled={deriveFromModuleRefernce === DERIVE_FROM_STEP_1.value}
+                            {...form.register('moduleReferenceString', {
+                                required: true,
+                                validate: validateModuleReference,
```
abizjak, line 414: Is this validation done on submission, or on every edit?

If it is the former we can make sure its 64 characters.

front-end-tools/src/writing_to_blockchain.ts
```diff
@@ -154,9 +156,9 @@ export async function initialize(
     let params: TypedSmartContractParameters | undefined;
 
     if (hasInputParameter) {
-        if (!useModuleFromStep1 && moduleSchema === undefined) {
+        if ((deriveFromModuleRefernce === DO_NOT_DERIVE.value || undefined) && moduleSchema === undefined) {
```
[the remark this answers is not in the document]
DOBEN: yes, I meant your last option but didn't write it in code :).

front-end-tools/src/components/InitComponent.tsx
```diff
@@ -95,11 +91,14 @@ export default function InitComponent(props: ConnectionProps) {
...
     const [schemaError, setSchemaError] = useState<string | undefined>(undefined);
 
-    const [isModuleReferenceAlreadyDeployedStep2, setIsModuleReferenceAlreadyDeployedStep2] = useState(false);
+    const [isModuleReferenceAlreadyDeployed, setIsModuleReferenceAlreadyDeployed] = useState(false);
 
     const [txHash, setTxHash] = useState<string | undefined>(undefined);
 
-    const [uploadedModuleSchemaBase64, setUploadedModuleSchemaBase64] = useState<string | undefined>(undefined);
+    const [schema, setSchema] = useState<string | undefined>(undefined);
+    const [moduleReference, setModuleReference] = useState<ModuleReference.Type | undefined>(
```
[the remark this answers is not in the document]
DOBEN: In case the user selects `deriveFromStep1` but the module reference is not set in step 1, we set the module reference back to `undefined` to catch that case. 
https://github.com/Concordium/concordium-smart-contract-tools/pull/154/files#diff-29a4ad93b8ac538e67ec59665a4623b87c298a4fc584824c3dc37430b233fe6dR327

front-end-tools/src/constants.ts
```diff
@@ -21,16 +21,33 @@ export const EXAMPLE_JSON_OBJECT = {
...
+    INPUT_PARAMETER_TYPE_ARRAY,
 ];
 
 // The subindex of all smart contracts.
 export const CONTRACT_SUB_INDEX = 0n;
 
-// Regular expression of a valid module reference which has to be a hex string `[0-9A-Fa-f]` of length 64.
-export const REG_MODULE_REF = /^[0-9A-Fa-f]{64}$/;
+// Characters of a hex string.
+export const hexRegex = /^[0-9a-fA-F]+$/;
```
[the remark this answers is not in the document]
DOBEN: I wanted to get rid of these hex expressions and rather use the following input `validation`. The input fields are immediately evaluated when the user fills them and an error appears at the front end in case of `wrong` input.

```
    const validateModuleReference = (value: string ) => {
        try {
        
            ModuleReference.fromHexString(value);

            return true;
        } catch (e) {
            return `Invalid module reference. Original error: ${(e as Error).message}`;
        }
    };
    
   ... 
   
   <Form.Control
        defaultValue={MODULE_REFERENCE_PLACEHOLDER}
        ...
        {...form.register('moduleReferenceString', {
               required: true,
              validate: validateModuleReference,
         })}
   />
```

But I observed that the method `ModuleReference.fromHexString(value)` only returns an error if the length of `value` does not match but not if the string characters are not hex. There is no evaluation if the string is actually a hex string since the type `HexString` in our SDK is just an alias for `String`.

```* @throws If the provided hex encoding does not correspond to a buffer of exactly 32 bytes.```
https://github.com/Concordium/concordium-node-sdk-js/blob/main/packages/sdk/src/types/ModuleReference.ts#L92
https://github.com/Concordium/concordium-node-sdk-js/blob/main/packages/sdk/src/types.ts#L35

Suggestion:
- Move back to regular expressions at this front end for now (checking length as well as hexCharacters). 
```
// Regular expression of a valid module reference which has to be a hex string `[0-9A-Fa-f]` of length 64.
export const REG_MODULE_REF = /^[0-9A-Fa-f]{64}$/;
```
- Add an evaluation to our `web-sdk` to check for if the type `HexString` consists of actual hex characters.

File: front-end-tools/src/components/InitComponent.tsx (hunk 3 above)
Comment (DOBEN) [reply front-end-tools/src/components/InitComponent.tsx:405]: It is done on every edit. The validation function checks both `lenght is 64 characters` and string is a hexstring.
https://github.com/Concordium/concordium-smart-contract-tools/pull/154#discussion_r1492449853

File: front-end-tools/src/components/InitComponent.tsx (hunk 1 above)
Comment (DOBEN) [reply front-end-tools/src/components/InitComponent.tsx]: It does check for the length in the function `ModuleReference.fromHexString(value)`
https://github.com/Concordium/concordium-node-sdk-js/blob/main/packages/sdk/src/types/ModuleReference.ts#L92

https://github.com/Concordium/concordium-smart-contract-tools/pull/154#discussion_r1492449853

**Review by abizjak — APPROVED** [review]
Ideally Soren would also have a look, but it looks OK to me.

File: front-end-tools/src/components/DeployComponent.tsx [hunk 7]
```diff
@@ -160,7 +143,7 @@ export default function DeployComponenet(props: ConnectionProps) {
 
                                 setBase64Module(module);
                                 setModuleReferenceCalculated(
-                                    Buffer.from(sha256([new Uint8Array(arrayBuffer)])).toString('hex')
+                                    ModuleReference.fromBuffer(Buffer.from(sha256([new Uint8Array(arrayBuffer)])))
```
soerenbf, line 146: Should be able to reduce this to

```suggestion
                                    ModuleReference.fromBuffer(sha256([new Uint8Array(arrayBuffer)]))
```

front-end-tools/src/components/InitComponent.tsx
```diff
@@ -142,29 +144,25 @@ export default function InitComponent(props: ConnectionProps) {
...
+                });
         }
     }, [connection, client, moduleReference]);
 
     const shouldWarnDifferenceModuleReferences = useMemo(() => {
         if (
             moduleReference !== undefined &&
             moduleReferenceCalculated !== undefined &&
-            moduleReferenceCalculated !== moduleReference
+            moduleReferenceCalculated.moduleRef !== moduleReference.moduleRef
```
soerenbf, line 165: Alternatively, consider using the following (which is generally available for all our wrapper types):
```suggestion
            ModuleReference.equals(moduleReferenceCalculated, moduleReference)
```

soerenbf (approved):
Looks good 😄

PR #162, opened 2024-03-25
Remove jenkins release job, add github action

abizjak:
## Purpose

Remove jenkins jobs for VSCode extension and add github action for it.

See https://github.com/Concordium/concordium-smart-contract-tools/actions/runs/8420891440 for a run of the release job without uploading artifacts to a release.

## Checklist

      hard-to-understand areas.


.github/workflows/release-vscode-extension.yml
```diff
@@ -0,0 +1,104 @@
...
+on:
+  # This release action always requires a tag to publish a release.
+  # This job will run when a releases/cargo-concordium/$VERSION
+  # tag is created.
+  push:
+    tags:
+      - releases/vscode-smart-contracts/*.*.*
+
+jobs:
+  build-cargo-concordium:
```
limemloh, line 18: **Discussion:** This is using cargo-concordium from the same commit rather than at some release tag. 

Maybe this is not an issue in practice, but since this is building cargo-concordium, the versioning could be a bit off if we don't release the extension right after releasing cargo-concordium.
Say that we release cargo-concordium, then fix a bug in cargo-concordium and then release the extension, this bug fix will be shipped with the extension.

limemloh (approved):

PR #167, opened 2024-05-14
Expose test module output path

lassemoldrup:
## Purpose

Addresses https://github.com/Concordium/concordium-rust-smart-contracts/issues/383.

## Changes

When running integration tests the module output path is now exposed to the tests via the `CARGO_CONCORDIUM_TEST_MODULE_OUTPUT_PATH` environment variable.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/cargo-concordium/src/build.rs
+++ b/cargo-concordium/src/build.rs
@@ -122,6 +122,8 @@ pub struct BuildInfo {
     /// list of file paths that were used to build the artifact. The file
     /// paths are relative to the package root.
     pub stored_build_info: Option<(utils::VersionedBuildInfo, Vec<PathBuf>)>,
+    /// The path to the file of the built module.
+    pub out_filename:      PathBuf,
 }
 
 /// Result of [`create_archive`]. It contains the actual archive with a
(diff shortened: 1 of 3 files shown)


cargo-concordium/src/build.rs
```diff
@@ -1202,6 +1206,13 @@ pub(crate) fn build_and_run_integration_tests(
 
         command.env("CARGO_CONCORDIUM_TEST_ALLOW_DEBUG", "1");
     }
+    // This enviroment variable needs to match the
+    // `CONTRACT_MODULE_OUTPUT_PATH_ENV_VAR` constant in the `contract-testing`
+    // crate.
+    command.env(
+        "CARGO_CONCORDIUM_TEST_MODULE_OUTPUT_PATH",
+        build_info.out_filename,
```
limemloh, line 1214: **Question:** Can we be sure this will always be an absolute path?

lassemoldrup: Why does it matter if it is absolute or relative?

limemloh: It might not matter, but I would be concerned that using a relative path would not work for every setup.

lassemoldrup: I don't think it should matter, since the `cargo test` process is spawned by `cargo-concordium`, so they should agree on paths in this case. I guess if we wanted to use the environment variable in other contexts it might matter, though.

limemloh (approved):

Issue #146, opened 2023-12-05
[front-end tools] Display raw bytes as hex string instead of buffer representation


lassemoldrup:
Suggestion: If we have no schema uploaded (and the read box returns the raw bytes), it would be better to return them e.g. in a hex type of string than the Buffer presentation (since for large objects it is hard to copy/read).

![Screenshot from 2023-11-29 13-24-43](https://github.com/Concordium/concordium-smart-contract-tools/assets/13039169/5742d9b0-381f-4cfc-aa0c-4a7a80b6a6e8)

_Originally posted by @DOBEN in https://github.com/Concordium/concordium-smart-contract-tools/issues/143#issuecomment-1831619130_

PR #169, opened 2024-06-12
Display raw bytes as hex string instead of a buffer

DOBEN:
## Purpose

closes https://github.com/Concordium/concordium-smart-contract-tools/issues/146

## Changes

- Display the raw bytes returned from the smart contract as hex string instead of buffer if no schema is provided.
- Distinguish between `errors` from the `ContractInvoke` and the `deriveFromContractIndex` flow. The later errors have no `errorCodes` so displaying links to error codes don't make sense for the later.

### DIFF
--- a/front-end-tools/src/reading_from_blockchain.ts
+++ b/front-end-tools/src/reading_from_blockchain.ts
@@ -251,9 +251,12 @@ export async function read(
         );
     }
 
+    // If no schema is provided, return the raw bytes as a hex string.
     if (moduleSchema === undefined) {
-        // If no schema is provided return the raw bytes as a hex string.
-        return JSONbig.stringify(`0x${uint8ArrayToHexString(res.returnValue.buffer)}`);
+        // Convert the Uint8Array to a hex string.
+        const hexString = uint8ArrayToHexString(res.returnValue.buffer);
+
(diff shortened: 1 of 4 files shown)


front-end-tools/src/reading_from_blockchain.ts
```diff
@@ -252,8 +252,8 @@ export async function read(
     }
 
     if (moduleSchema === undefined) {
-        // If no schema is provided return the raw bytes
-        return JSONbig.stringify(res.returnValue);
+        // If no schema is provided return the raw bytes as a hex string.
+        return JSONbig.stringify(`0x${uint8ArrayToHexString(res.returnValue.buffer)}`);
```
Radiokot, line 256: So it will return a double-quoted string? Why double quotes are needed?

Radiokot (approved):

front-end-tools/src/reading_from_blockchain.ts
DOBEN: good point, a pure `a double-quoted string` doesn't look good (and there is no information on what this string represents).

I return an object now that has a human-readable field name `rawBytes` instead as follows:
Screenshot from 2024-06-13 11-28-28

PR #176, opened 2024-06-27
Change default build output path

### LINKED ISSUE #155: Cargo-concordium: output to `concordium-out/module.wasm.v1` by default
**Description**
Cargo concordium should output the wasm module to `PROJECT_ROOT/concordium-out/module.wasm.v1` by default, as it is cumbersome and error-prone to write manually all the time. Where `PROJECT_ROOT` is the folder with the manifest file (`Cargo.toml`).
All of our example contracts with integration tests require this location anyhow and we also recommend it in our developer documentation.

The only downside is that it is a breaking change.

lassemoldrup:
## Purpose

Closes #155.

## Changes

Changes the default output to `concordium-out/module.wasm.v1` (or `.v0`). Also removes the requirement for `--out`, when using `--verifiable`.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/cargo-concordium/src/build.rs
+++ b/cargo-concordium/src/build.rs
@@ -414,9 +414,9 @@ pub(crate) fn build_contract(
 
     let package_version_string = format!("{}-{}", package.name, package.version);
 
-    // If the output path is supplied make sure up-front before building anything
-    // that it points to a sensible location
-    let out_filename = match &out {
+    // Make sure up-front before building anything that the output path points to a
+    // sensible location
+    let mut out_filename = match &out {
         Some(out) => {
             // A path and a filename need to be provided when using the `--out` flag.
             if out.file_name().is_none() || out.is_dir() {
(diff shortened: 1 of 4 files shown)


cargo-concordium/src/build.rs
```diff
@@ -397,20 +397,49 @@ pub(crate) fn build_contract(
...
+    let mut out_filename = match &out {
+        Some(out) => {
+            // A path and a filename need to be provided when using the `--out` flag.
+            if out.file_name().is_none() || out.is_dir() {
+                anyhow::bail!(
+                    "The `--out` flag requires a path and a filename (expected input: \
+                     `./my/path/my_smart_contract.wasm.v1`)"
+                );
+            }
+            out.clone()
```
DOBEN, line 428:
```suggestion
 let mut out_filename = match out {
        Some(out) => {
            // A path and a filename need to be provided when using the `--out` flag.
            if out.file_name().is_none() || out.is_dir() {
                anyhow::bail!(
                    "The `--out` flag requires a path and a filename (expected input: \
                     `./my/path/my_smart_contract.wasm.v1`)"
                );
            }
            out
        }
```

DOBEN (approved):

limemloh (approved):

cargo-concordium/src/build.rs
lassemoldrup: Good catch!

PR #177, opened 2024-06-27
Embed schema by default

### LINKED ISSUE #173: Cargo-concordium: embed schema by default
**Description**
We should consider having `cargo-concordium` embed the schema by default when building smart contracts, since it is most often what you want. If we do this, we need to add a flag for opting out. This could be `--no-schema-embed`/`-n`. Maybe it is worth keeping the existing `--schema-embed`/`-e` flag as a (deprecated) no-op to not break people's working build commands?

lassemoldrup:
## Purpose

Closes #173. The branch is branched out from #176.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/cargo-concordium/CHANGELOG.md
+++ b/cargo-concordium/CHANGELOG.md
@@ -7,7 +7,7 @@
   tests via the `CARGO_CONCORDIUM_TEST_MODULE_OUTPUT_PATH` environment variable.
 - Change the default build output path to `concordium-out/module.wasm.v1`.
 - Remove requirement for `--out` flag when using the `--verifiable` flag.
-- Embed the schema in the Wasm module by default. This can be disabled with the `--no-schema-embed` flag.
+- Embed the schema in the Wasm module by default. Therefore, the `--schema-embed` flag is now deprecated. This behavior can be disabled with the `--no-schema-embed` flag.
 
 ## 3.3.0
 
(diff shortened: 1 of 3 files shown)


cargo-concordium/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 - Added a `tag` option to `cargo concordium init`
 - When running integration tests the module output path is now exposed to the
   tests via the `CARGO_CONCORDIUM_TEST_MODULE_OUTPUT_PATH` environment variable.
+- Change the default build output path to `concordium-out/module.wasm.v1`.
+- Remove requirement for `--out` flag when using the `--verifiable` flag.
+- Embed the schema in the Wasm module by default. This can be disabled with the `--no-schema-embed` flag.
```
DOBEN, line 10: Please mention that the `--schema-embed` flag has been deprecated as a result.

cargo-concordium/src/main.rs
```diff
@@ -1113,6 +1122,16 @@ fn handle_build(options: BuildOptions, print_extra_info: bool) -> anyhow::Result
             bold_style.paint(size)
         );
     }
+    if options.schema_embed {
+        let error_style = ansi_term::Color::Yellow;
+        eprintln!(
+            "{}",
+            error_style.paint(
+                "\n\nThe `--schema-embed`/`-e` option is no longer necessary, as it is now the \
```
DOBEN, line 1130:
```suggestion
                "\n\nThe `--schema-embed`/`-e` flag is no longer necessary, as it is now the \
```

DOBEN (approved):
Eventually, we should update our developer documentation and remove the deprecated flag from the commands. Just not immediately since most people will still be using older versions of `cargo-concordium`.

limemloh (approved):

PR #186, opened 2024-10-28
Build cargo-concordium for Mac ARM64 in vscode release

limemloh:
## Purpose

Our vscode extension distribution for MacOS ARM is shipping with `cargo-concordium` build for `x86_64` which means the user depends on rosetta.
This PR will add cross-compiling for Macos and prepare a new release with this binary.

### DIFF
--- a/.github/workflows/release-vscode-extension.yml
+++ b/.github/workflows/release-vscode-extension.yml
@@ -36,15 +36,26 @@ jobs:
         run: |
           rustup target add x86_64-unknown-linux-musl
           cargo build --release --manifest-path cargo-concordium/Cargo.toml --target x86_64-unknown-linux-musl
-      - name: Build on other platforms
-        if: matrix.platfrom != 'ubuntu-latest'
+      - name: Build on windows
+        if: matrix.platfrom == 'windows-latest'
         run: cargo build --release --manifest-path cargo-concordium/Cargo.toml
+      - name: Build on macos
+        if: matrix.platfrom == 'macos-latest'
+        run: |
+          # Build x86_64
+          cargo build --release --manifest-path cargo-concordium/Cargo.toml
+          # Build ARM64 and rename it.
+          rustup target add aarch64-apple-darwin
+          cargo build --release --manifest-path cargo-concordium/Cargo.toml --target aarch64-apple-darwin
+          mv cargo-concordium/target/release/aarch64-apple-darwin/cargo-concordium cargo-concordium/target/release/aarch64-apple-darwin/cargo-concordium-aarch64
+
       - name: Upload artifact
         uses: actions/upload-artifact@v4
         with:
           name: binary-${{ matrix.platform }}
           path: |
             cargo-concordium/target/release/cargo-concordium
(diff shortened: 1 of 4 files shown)


.github/workflows/release-vscode-extension.yml
```diff
@@ -36,15 +36,26 @@ jobs:
...
         run: cargo build --release --manifest-path cargo-concordium/Cargo.toml
+      - name: Build on macos
+        if: matrix.platfrom == 'macos-latest'
+        run: |
+          # Build x86_64
+          cargo build --release --manifest-path cargo-concordium/Cargo.toml
+          # Build ARM64 and rename it.
+          rustup target add aarch64-apple-darwin
+          cargo build --release --manifest-path cargo-concordium/Cargo.toml --target aarch64-apple-darwin
+          mv cargo-concordium/target/release/aarch64-apple-darwin/cargo-concordium cargo-concordium/target/release/aarch64-apple-darwin/cargo-concordium-aarch64
```
soerenbf, line 50: It seems redundant to specify architecture in both the directory and filename?

```suggestion
```

Alternatively move the file to 
```
cargo-concordium/target/release/cargo-concordium-aarch64
```

.github/workflows/release-vscode-extension.yml
```diff
@@ -36,15 +36,26 @@ jobs:
...
+
       - name: Upload artifact
         uses: actions/upload-artifact@v4
         with:
           name: binary-${{ matrix.platform }}
           path: |
             cargo-concordium/target/release/cargo-concordium
+            cargo-concordium/target/release/aarch64-apple-darwin/cargo-concordium-aarch64
```
soerenbf, line 58: And then

```suggestion
            cargo-concordium/target/release/aarch64-apple-darwin/cargo-concordium
```

.github/workflows/release-vscode-extension.yml
```diff
@@ -68,11 +79,15 @@ jobs:
...
-          # Mac first
+          # Mac x86_64 first
           mv ../binary-macos-latest/cargo-concordium executables/cargo-concordium
           chmod +x executables/cargo-concordium
-
           npx vsce package --target darwin-x64 --out ./out/extension-darwin-x64.vsix
+
+          # Then Mac ARM64
+          rm -rf executables/*
+          mv ../binary-macos-latest/cargo-concordium-aarch64 executables/cargo-concordium
```
soerenbf, line 89: This can't be right? When specifying multiple paths for `action/upload-artifact` the least common ancestor directory is used as the root ([docs](https://github.com/actions/upload-artifact?tab=readme-ov-file#upload-using-multiple-paths-and-exclusions)). That would mean this

```suggestion
          mv ../binary-macos-latest/aarch64-apple-darwin/cargo-concordium-aarch64 executables/cargo-concordium
```

.github/workflows/release-vscode-extension.yml
limemloh: This is to avoid the name clashing with the `x86_64` build as they get uploaded as artifacts.

soerenbf (approved):

PR #197, opened 2025-02-27
Bumped version and updated base

rasmus-kirk:
## Purpose

See:

https://github.com/Concordium/concordium-base/pull/563

### DIFF
--- a/cargo-concordium/CHANGELOG.md
+++ b/cargo-concordium/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
-## Unreleased changes
+## 4.1.0
+
+- Updated concordium-base dependency
 
 ## 4.0.0
 
--- a/cargo-concordium/CHANGELOG.md
+++ b/cargo-concordium/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 4.2.0
 
+- Unit tests now run in parallel and the results are printed sequentially.
 - Updated concordium-base dependency
   - The cargo concordium test now allow tests to use several getter and setter functions in unit tests.
 
--- a/cargo-concordium/CHANGELOG.md
+++ b/cargo-concordium/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 4.1.0
+## 4.2.0
 
 - Updated concordium-base dependency
 
--- a/cargo-concordium/CHANGELOG.md
+++ b/cargo-concordium/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 4.2.0
 
 - Updated concordium-base dependency
-
+  - The cargo concordium test now allow tests to use several getter and setter functions in unit tests.
 ## 4.1.0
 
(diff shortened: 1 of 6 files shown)


limemloh (approved):

cargo-concordium/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Changelog
 
-## Unreleased changes
+## 4.2.0
+
+- Updated concordium-base dependency
```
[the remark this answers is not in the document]
rasmus-kirk: Will do

cargo-concordium/src/build.rs
```diff
@@ -1265,6 +1269,72 @@ pub(crate) fn build_and_run_integration_tests(
...
+    match test_result {
+        Some(err) => {
+            is_err = true;
+            eprintln!(
+                "  - {} ... {}",
+                test_name,
+                Color::Red.bold().paint("FAILED")
+            );
+            eprintln!(
+                "    {} ... {}",
```
limemloh, line 1309: Since this function is run in parallel, the different `eprintln!` might be intertwined right?
Maybe we should print these in sequence and only have the artifact.run be done in parallel.
What do you think?

rasmus-kirk: I see the issue, but won't moving this out of the parallelized section mean that the code will wait for all artifacts to finish before printing? I don't mind refactoring the implementation, but I would like to retain that functionality. I'll try something, until you come on again.

rasmus-kirk: How's this?

limemloh: :+1: Great approach

cargo-concordium/src/main.rs
```diff
@@ -684,9 +692,15 @@ pub fn main() -> anyhow::Result<()> {
         Command::Test {
             seed,
             build_options,
+            test_threads,
             only_unit_tests,
             test_targets,
         } => {
+            if let Some(test_threads) = test_threads {
```
rasmus-kirk, line 699: In the previous commit, this if statement was missing. This still worked as passing the default `0` to rayon is the same as not calling this code at all. However, since this code could fail if `.build_global()` in the future is called twice at runtime, I would rather have it only fail if a flag is specifically passed.

cargo-concordium/src/main.rs
```diff
@@ -87,6 +87,13 @@ This command also builds a deployable Wasm module for integration testing, and i
             help = "Whether only the unit tests should be run."
         )]
         only_unit_tests: bool,
+        #[structopt(
+            name = "test-threads",
+            long = "test-threads",
+            help = "How many threads should be used for running unit tests. Defaults all \
+                    available threads."
```
limemloh, line 94:
```suggestion
            help = "How many threads should be used for running unit tests. Defaults to the number of logical CPUs."
```

limemloh (approved):

PR #201, opened 2025-06-26
COR-1515: Use wasm-opt in verified smart contracts

rasmus-kirk:
## Purpose

Fixes broken smart contract building for verified smart contracts. There's no need to update changelogs as the docker image doesn't have one, and the cargo concordium changes are mostly just me expanding on documentation/errors I didn't grok initially.

### DIFF
--- a/.github/workflows/release-reproducible-smart-contract.yaml
+++ b/.github/workflows/release-reproducible-smart-contract.yaml
@@ -42,11 +42,11 @@ jobs:
             CARGO_CONCORDIUM="cargo run --manifest-path $GITHUB_WORKSPACE/cargo-concordium/Cargo.toml --target x86_64-unknown-linux-gnu -- concordium"
             $CARGO_CONCORDIUM build --verifiable "concordium/$SERVICE_NAME:$RUST_VERSION"
             $CARGO_CONCORDIUM verify-build --module ./concordium-out/module.wasm.v1 --source ./concordium-out/module.wasm.v1.tar
-      # - name: Log in to the Container registry
-      #   uses: docker/login-action@65b78e6e13532edd9afa3aa52ac7964289d1a9c1
-      #   with:
-      #     registry: ${{ env.REGISTRY }}
-      #     username: ${{ secrets.DOCKERHUB_USERNAME }}
-      #     password: ${{ secrets.DOCKERHUB_TOKEN }}
-      # - name: Push the image
-      #   run: docker push "concordium/$SERVICE_NAME:$RUST_VERSION"
+      - name: Log in to the Container registry
+        uses: docker/login-action@65b78e6e13532edd9afa3aa52ac7964289d1a9c1
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Push the image
+        run: docker push "concordium/$SERVICE_NAME:$RUST_VERSION"


cargo-concordium/src/build.rs
```diff
@@ -156,10 +156,7 @@ fn create_archive(
     tar.mode(tar::HeaderMode::Deterministic);
     // Ignore files that are ignored by Git.
     let files = ignore::WalkBuilder::new(package_root_path)
-        .git_global(true)
-        .git_ignore(true)
-        .parents(true)
-        .hidden(true)
+        .standard_filters(true)
```
limemloh, line 159: Not that I am against this, but why the change?

This should also be reflected in the changelog

reproducible/build.Dockerfile
```diff
@@ -5,11 +5,12 @@
...
 
-ARG source_image
-FROM ${source_image} AS build
+FROM rust:1.85 AS build
 
 RUN rustup target add wasm32-unknown-unknown
 
 RUN mkdir /b
 
+RUN cargo install wasm-opt --version 0.116.1
```
limemloh, line 14: To ensure this is somewhat reproduceable, it should use locked dependencies
```suggestion
RUN cargo install wasm-opt --locked --version 0.116.1
```

cargo-concordium/src/main.rs
```diff
@@ -313,11 +313,16 @@ struct VerifyOptions {
...
+        help = "Path to the source code of the code, this would be the tar produced during the \
+                reproducible build. If not present then the sources will be downloaded from an \
                 embedded link in the build info."
     )]
     source_path:       Option<PathBuf>,
-    #[structopt(name = "module", long = "module", help = "Module to verify.")]
+    #[structopt(
+        name = "module",
+        long = "module",
+        help = "Path to the module WASM file to verify."
```
limemloh, line 324: Concordium smart contract modules are actually not valid Wasm modules... we prefix a few bytes, making them invalid, so maybe:
```suggestion
        help = "Path to the smart contract module file to verify."
```

cargo-concordium/src/build.rs
rasmus-kirk: Motivation was that it seemed like it was what the current code was trying to do. But given that this is affects the changelog, I would rather keep this as is.

reproducible/build.Dockerfile
rasmus-kirk: Good catch

allanbrondum (approved):

.github/workflows/release-reproducible-smart-contract.yaml
```diff
@@ -10,7 +10,6 @@ on:
 env:
   REGISTRY: docker.io
   SERVICE_NAME: 'verifiable-sc'
-  SOURCE_IMAGE_TAG: 'rust:1.75.0'
```
DOBEN, line 13: The env variable `SOURCE_IMAGE_TAG` is still used in this workflow so that needs some clean-up
  
  ```     
            if [ $SERVICE_TAG != ${{ env.SOURCE_IMAGE_TAG }} ]; then
                echo "Tag version does not match the source image version."
                exit 1
            fi
   ```

.github/workflows/release-reproducible-smart-contract.yaml
```diff
@@ -47,5 +46,3 @@ jobs:
           push: true
           platforms: linux/amd64
           tags: ${{ steps.meta.outputs.tag }}
-          build-args: |
-            source_image=${{ env.SOURCE_IMAGE_TAG }}
```
DOBEN, line 51: The idea is that we ensure that the docker `tag` matches the rust version inside of the docker build. Currently, we not ensuring this anymore. 

Meaning if I make a realize with the below CI pipeline tag, then the rust version in the docker images should be `1.87.0` and not the hardcoded one `1.85`

```
docker pull concordium/verifiable-sc:1.87.0
```

reproducible/build.Dockerfile
```diff
@@ -5,11 +5,12 @@
 # to copy data in and out of the container. The image also adds
 # wasm32-unknown-unknown target.
 
-ARG source_image
-FROM ${source_image} AS build
+FROM rust:1.85 AS build
```
DOBEN, line 8: We want to build the docker image for different rust versions.

.github/workflows/release-reproducible-smart-contract.yaml
allanbrondum: Yes, I think some validation was lost by the changes. I look primarily at the changes since Emils review, so did not see it, but yes. Seems like now the CI does not verify the tag matches the Rust version.

allanbrondum: Though the release build will likely fail hard right now, since the validation code is still there :)

.github/workflows/release-reproducible-smart-contract.yaml
```diff
@@ -21,31 +21,32 @@ jobs:
...
+          toolchain: ${{ env.RUST_VERSION }}
+          target: wasm32-unknown-unknown
+      - run: docker build -f ./reproducible/build.Dockerfile --build-arg source_image="rust:$RUST_VERSION" -t "concordium/$SERVICE_NAME:$RUST_VERSION" ./reproducible
+      - name: Test smart contract
+        working-directory: ./reproducible/example-contract
+        run: |
+            CARGO_CONCORDIUM="cargo run --manifest-path $GITHUB_WORKSPACE/cargo-concordium/Cargo.toml --target x86_64-unknown-linux-gnu -- concordium"
+            $CARGO_CONCORDIUM build --verifiable "concordium/$SERVICE_NAME:$RUST_VERSION"
+            $CARGO_CONCORDIUM verify-build --module ./concordium-out/module.wasm.v1 --source ./concordium-out/module.wasm.v1.tar
       - name: Extract version from git tag
```
limemloh, line 35: Maybe move this step up and check the version from the tag before building the image, since it would fail a lot faster then.

reproducible/build.Dockerfile
```diff
@@ -1,15 +1,17 @@
 # A builder for images for verifiable/reproducible builds.
-# The `source_image` is meant to be a published Rust image, such as rust:1.70.
+# The `source_image` is meant to be a published Rust image, such as rust:1.90.
 #
 # This image only adds a script `run-copy.sh` that is used by cargo-concordium
 # to copy data in and out of the container. The image also adds
 # wasm32-unknown-unknown target.
 
-ARG source_image
+ARG source_image=rust:1.86.0
```
limemloh, line 8: I would prefer if the flow fails instead of having a default source image here.

.github/workflows/release-reproducible-smart-contract.yaml
```diff
@@ -21,31 +21,32 @@ jobs:
...
-          registry: ${{ env.REGISTRY }}
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_TOKEN }}
+          toolchain: ${{ env.RUST_VERSION }}
+          target: wasm32-unknown-unknown
+      - run: docker build -f ./reproducible/build.Dockerfile --build-arg source_image="rust:$RUST_VERSION" -t "concordium/$SERVICE_NAME:$RUST_VERSION" ./reproducible
+      - name: Test smart contract
+        working-directory: ./reproducible/example-contract
+        run: |
+            CARGO_CONCORDIUM="cargo run --manifest-path $GITHUB_WORKSPACE/cargo-concordium/Cargo.toml --target x86_64-unknown-linux-gnu -- concordium"
```
limemloh, line 32:
```suggestion
            CARGO_CONCORDIUM="cargo run --manifest-path $GITHUB_WORKSPACE/cargo-concordium/Cargo.toml -- concordium"
```
Is the target needed here?

limemloh (commented):
Looks good with the test.
It would be good to document how to release for a new rust version in a README of the reproducible directory.

.github/workflows/release-reproducible-smart-contract.yaml
limemloh: ahh right... so this was the issue we talked about earlier.
Okay, then leave it in

